PR #5, opened 2026-03-16
Rewrite WakeModule with file-based policies and debounce

antra-tess:
## Summary

- **File-based wake policies**: WakeModule rewritten from tool-driven subscriptions to a `wake.json` config file. Recipes seed the config on first run; the agent can edit it at runtime with file tools.
- **Per-policy debounce**: Three behaviors — `"always"` (trigger immediately), `"suppress"` (drop), `{ "debounce": ms }` (batch per-policy, deliver when timer fires)
- **Structural matching**: Policies match on `scope` (event type), `source` (serverId glob), `channel` (channelId glob), and `filter` (text/regex on content)
- **Recipe integration**: `modules.wake` accepts `WakeConfig` object to seed `wake.json` — zulip-miner recipe updated as example

### Performance
- Compiled regex/glob cache (built once at config load, not per-event)
- Throttled filesystem checks (1s interval for hot-reload)
- Inference buffer capped at 100 events
- Deferred flush via `queueMicrotask` to avoid trace callback re-entrancy

### Companion change
- agent-framework `feature/mcpl-v05-state-branches`: enriches `shouldTriggerInference` metadata with `serverId`, `channelId`, `author` so policies can match on channel structure

## Test plan

- [x] 27 unit tests covering policy matching, debounce timing, inference buffering, config hot-reload, and seeding
- [ ] `bun src/index.ts recipes/zulip-miner.json` — wake.json seeded from recipe
- [ ] Edit `wake.json` at runtime → policies reload within 1s
- [ ] Debounced events batch correctly and deliver after timer fires

### DIFF
--- a/recipes/mcpl-editor-test.json
+++ b/recipes/mcpl-editor-test.json
@@ -5,7 +5,7 @@
 
   "agent": {
     "name": "tester",
-    "model": "claude-sonnet-4-6-20250514",
+    "model": "claude-sonnet-4-6",
     "maxTokens": 8192,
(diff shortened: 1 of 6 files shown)


Anarchid:
Superceded by the Event Gate framework-level feature

PR #14, opened 2026-04-22
feat(fleet): onIdle hook for debounced auto-command dispatch

Tengro:
## Summary

Adds a per-child `onIdle` hook to `FleetModule` that auto-dispatches a slash command when a child's `lifecycle:idle` event persists for a configured debounce period. Use case that drove this: ticket-driven miners in the Knowledge Mining Triumvirate — between tickets, the miner's context should compress (`/newtopic`) automatically so the next ticket starts with a fresh head window, without needing the conductor agent to watch idle events and dispatch manually.

### `FleetModule.onIdle`
```jsonc
"children": [{
  "name": "miner",
  "recipe": "recipes/knowledge-miner.json",
  "onIdle": {
    "command": "/newtopic",
    "debounceMs": 120000
  }
}]
```

Semantics:
- On `lifecycle:idle` from the child → arm a `debounceMs` timer.
- Any subsequent event from the child → cancel the pending timer.
- Timer elapses → dispatch the configured slash command via the same `sendToChild` path `fleet--command` uses.
- Re-arms on every subsequent idle, so the hook fires once per sustained-idle period rather than on every between-round transient idle.
- Default `debounceMs`: 120_000 (2 minutes).

### Infinite-loop guard in the child runtime

Naively, `/newtopic` dispatch would create a loop: auto-`/newtopic` → `/newtopic` runs an LLM summary → child goes idle again → another `lifecycle:idle` fires → timer arms → another `/newtopic`, and so on over a near-empty head.

Fix: the child's headless runtime resets its idle-tracking state (`hadAtLeastOneInference`, `idleSince`, `idleEmitted`, `currentSpeech`) whenever `/newtopic` runs. No fresh `inference:started` = no new `lifecycle:idle`. The reset is wired through `dispatchCommand` in `src/headless.ts` so it covers both onIdle-dispatched and operator-typed `/newtopic`.

This also composes cleanly with the pre-existing invariant that boot-and-sit-idle doesn't emit `lifecycle:idle` until a real inference has happened once — so the hook can't fire on a brand-new child sitting quietly waiting for its first ticket.

### Triumvirate recipe wiring

`recipes/triumvirate.json` now wires `onIdle: { command: "/newtopic", debounceMs: 120000 }` on the miner as the demonstrator. The clerk and reviewer don't need it — the clerk's work is event-bounded (reply-or-ticket, then done), the reviewer's is file-driven (one pass per new library doc, then done). The miner is the only child whose task spans many inferences and benefits from explicit compression between tasks.

### Recipe schema

`RecipeFleetChild.onIdle?: { command, debounceMs? }` added to the tracked types. Validation enforces:
- `onIdle` is an object (if present)
- `command` is a non-empty string
- `debounceMs` is a positive number (if present)

Out-of-shape configs fail at recipe load, not at runtime.

### Timer-lifetime hygiene

- `proc.on('exit')` clears any pending `onIdleTimer` (no dangling setTimeout after the child dies).
- `FleetModule.stop()` iterates and clears before kill broadcast.
- `killChild()` path is covered by the `exit` handler.
- Adopt-on-restart (parent re-launch reattaching to a live child) sets `onIdle: null` on the reconstructed orphan — the original timer state isn't meaningful after reattach; first real idle post-reattach re-arms fresh against the new parent's config.

### `.gitignore`

Added `knowledge-requests/` — runtime ticket directory (clerk writes here), parallel to existing `output/` and `review-output/`. Noticed during local triumvirate testing; not strictly part of the onIdle feature but tightly coupled operationally.

## Test plan

92 tests total (4 new), all passing.

### New tests — `test/fleet-onidle.test.ts`

- [x] **fires configured command after sustained idle** — launch mock child with `onIdle: { command: "/help", debounceMs: 300 }`, send a text to drive an inference + idle, wait past the debounce, observe `/help` arrived at the child (mock echoes received commands as `command-output` events).
- [x] **activity during debounce cancels + re-arms** — launch with debounce 500ms, send first text (armed), wait 200ms, send second text (should cancel pending, re-arm on new idle), wait 400ms (< debounce from new arm → no dispatch yet), assert no `/help` arrived, wait another 500ms (past the new arm's debounce), assert `/help` did arrive.
- [x] **schema validation** — `validateRecipe` rejects non-object `onIdle`, missing/empty `command`, non-positive `debounceMs`; accepts well-formed.
- [x] **pending timer cleared on child kill** — arm a 5-second timer, send the `kill` tool, assert `child.onIdleTimer === null` (test harness would hang waiting for orphaned `setTimeout` otherwise).

### Mock child

`test/mock-headless-child.ts` extended to echo any unrecognized slash command as `command-output: mock-received-command: <cmd>` so the test can observe onIdle arrivals without needing the command to be meaningfully implemented in the mock.

## Manual verification for operators

After merging, update the triumvirate setup:

1. First run of the day, miner gets woken by a new ticket in `knowledge-requests/`.
2. Miner does its mining, writes docs to `library-mined/`.
3. Miner goes idle.
4. 2 minutes later (no new tickets) → `/fleet peek miner` should show a `command-output` line from the auto-dispatched `/newtopic`.
5. Next ticket arrives later that day → miner works with a fresh head window, no carryover from the prior ticket polluting its context.

If the miner's context feels too tight (you want more carryover between similar tickets), bump `debounceMs` on the triumvirate.json entry. If you want to disable the hook entirely for a given run, delete the `onIdle` block from the recipe.

## Out of scope

- Other event hooks (`onCrash`, `onReady`, `onToolFailure`). `onIdle` was the concrete need. If new ones land, the same shape (`{ event: "...", command, debounceMs? }`) would extend cleanly.
- Text-message dispatch (as opposed to slash-command). Deliberate: slash commands are structured and idempotent; text messages trigger inference and would compound the noise problem onIdle is supposed to reduce.
- TUI-path reset of idle tracking. The TUI's internal state machine doesn't use the headless `hadAtLeastOneInference` flag, so nothing to reset there — the equivalent happens naturally via `refreshFromStore` on branch changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ mcpl-servers.json
 my-recipe.json
 output/
 review-output/
+knowledge-requests/
 tsconfig.tsbuildinfo
--- a/recipes/triumvirate.json
+++ b/recipes/triumvirate.json
@@ -29,7 +29,11 @@
           "recipe": "recipes/knowledge-miner.json",
           "dataDir": "./data/miner",
           "autoStart": true,
-          "subscription": ["lifecycle", "inference:completed", "inference:speech", "tool:completed", "tool:failed", "inference:failed"]
+          "subscription": ["lifecycle", "inference:completed", "inference:speech", "tool:completed", "tool:failed", "inference:failed"],
+          "onIdle": {
+            "command": "/newtopic",
+            "debounceMs": 120000
+          }
         },
         {
           "name": "reviewer",
--- a/src/headless.ts
+++ b/src/headless.ts
@@ -142,6 +142,14 @@ export async function runHeadless(app: AppContext, argv: string[] = []): Promise
     emit(traceEvent as unknown as Record<string, unknown>);
   });
 
+  // Filled in by the idle-tracking block below.  Called whenever an event
+  // has semantically reset the conversation head (currently: /newtopic).
+  // After a reset, the poll must wait for a fresh `inference:started`
+  // before emitting `lifecycle:idle` again — otherwise a post-compression
+  // idle would instantly trigger whatever auto-command is watching for
+  // idle (e.g. an onIdle hook that fires /newtopic) in an infinite loop.
+  let resetIdleTracking: () => void = (): void => { /* wired below */ };
+
   // -- Command dispatch --
   async function dispatchCommand(cmd: IncomingCommand): Promise<void> {
     switch (cmd.type) {
@@ -181,6 +189,14 @@ export async function runHeadless(app: AppContext, argv: string[] = []): Promise
         if (result.switchToSessionId) {
           await app.switchSession(result.switchToSessionId);
           emit({ type: 'command-output', text: 'Session switched.', style: 'system' });
+        }
+        // `/newtopic` compresses the head window; the head is now logically
+        // fresh.  Clear the idle-tracker so a subsequent lifecycle:idle
+        // won't fire until after the next real inference.  Relevant even
+        // outside onIdle-hook scenarios: any operator running /newtopic
+        // interactively gets the same well-defined post-reset state.
+        if (/^\/newtopic(?:\s|$)/.test(cmd.command)) {
+          resetIdleTracking();
         }
         if (result.quit) {
           await gracefulShutdown('command:/quit');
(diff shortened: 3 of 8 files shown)


Anarchid:
(closing due to uncertain utility, will revisit the issue of `/newtopic` and sticky headers with more focus)

PR #21, opened 2026-05-06
feat(webui): admin SPA + JSON-over-WS control plane

Anarchid:
## Summary

Adds a web admin UI for connectome-host: a single-page app (Solid + Vite + Tailwind) served alongside a JSON-over-WebSocket control plane, mirroring what TuiModule provides for the terminal — message stream, agent tree, slash commands, peek panes for subagents and fleet children, plus admin panels for lessons, MCPL servers, and workspace files.

Architectural notes:
- **Single fan-out trace listener.** The framework's `onTrace` doesn't return a detacher, so we keep one listener and let the WS lifecycle own client membership.
- **Process-level singleton server.** Sessions rebuild the framework, but the open WebSocket connections stay up across session switches — modules rebind their AppContext on `setApp()` instead of restarting the server.
- **Wire protocol is versioned** (`WEB_PROTOCOL_VERSION = 0`); clients refuse to connect on mismatch.
- **TLS / outer auth is the reverse proxy's job.** Module ships with optional Basic-Auth as defense-in-depth and refuses to bind a non-loopback host without `basicAuth` or an explicit `acknowledgeNoAuth: true`. See `docs/webui-deployment.md` for the recommended Caddy front.
- **Fleet-scoped panels.** Lessons / Workspace / Usage / Peek all support a per-child scope via `corrId`-routed IPC through the existing FleetModule.

## Security hardening (last commit)

A QA pass surfaced two emergency-grade issues; both are fixed in `620716c`:

- **XSS in markdown rendering.** marked v14 has no built-in sanitizer and the assistant turn's text was piped straight to `innerHTML`. A tool result, peeked subagent line, or mined-knowledge ingestion containing literal HTML would land as live JS in the operator's browser. Now sanitized through DOMPurify.
- **WebSocket CSRF on `/ws`.** No Origin check. Browsers do not enforce same-origin on `new WebSocket(...)`, so any tab the operator opened could drive the localhost-bound admin. Added an Origin allowlist (default loopback; configurable via `RecipeWebUi.allowedOrigins`) checked before auth on upgrade.

Plus five concerning items:
- Basic-auth compare is constant-time (SHA-256 + `timingSafeEqual`).
- `isClientMessage` narrows per variant; `mcpl-add` rejects non-string fields, oversized ids, path separators, control chars.
- `pendingFleetRequests` actually prunes by TTL and sends a synthetic timeout to the originating client (previously the TTL field was written but never read).
- Workspace file reads cap at 256 KB *and* 5000 lines, truncating at a UTF-8 boundary.
- `serveStatic` path containment now requires `root + sep`, defeating the sibling-prefix bypass (`<root>-evil/...`).

## Test plan

- [x] `bun test` — 205 pass / 0 fail (32 new; 22 protocol-validation, 10 HTTP/WS smoke)
- [x] `bunx tsc --noEmit` — only the documented pre-existing `MessageQuery`/CM-duplicate errors
- [x] `bun run build` (web) — 19 modules, 138 KB / 42 KB gzip
- [x] Local smoke: open the SPA in Firefox, route a message via `@miner`, peek a fleet child, hit /quit with children running and exercise the modal, edit an MCPL server entry from the admin panel
- [ ] Reviewer: confirm Caddy/proxy deployment still works with the new default Origin allowlist (override via `allowedOrigins` if your proxy rewrites the Host)
- [ ] Reviewer: confirm the markdown rendering still looks right after DOMPurify

## Follow-ups (not in this PR)

The QA review also flagged polish items deferred for separate PRs:
- Symbol-keyed framework property → put on `AppContext` like every other module hook
- Split the 1675-line `web-ui-module.ts` (HTTP, WS handlers, fleet bridge, welcome assembly)
- Split the 1245-line `App.tsx` (Header/MessageView/CommandSuggestions/RecipePane/SidebarTabs into their own files)
- Memoize repeated `getAllModules().find(...)` lookups behind typed accessors on AppContext
- Tool-call rendering should expose input on hover/expand; currently shows only the bare name


Anarchid:
image

PR #22, opened 2026-05-10
feat(host): persistent LLM-call log via membrane.beforeRequest hook

antra-tess:
## Summary

Writes one JSON line per LLM request to \`{dataDir}/llm-calls.jsonl\`. Each entry: \`{ ts, normalizedConfig, rawRequest }\`.

## Why

Nothing in the agent-framework / connectome-host stack persisted the actual LLM request body anywhere. When the API returned 4xx with a flash error in the TUI, post-mortem inspection was impossible.

Caught in the wild during a live Lena bring-up: a stale \`agent-framework/node_modules/@animalabs/context-manager\` (real directory shadowing a linked source) was injecting \`temperature: 0\` into compression calls, which \`claude-opus-4-7\` rejects. Without this log, diagnosis would have required instrumenting across 4 files in 3 packages.

## How

Uses the existing \`MembraneConfig.hooks.beforeRequest\` contract — receives both the normalized request (what the caller built) and the provider-format raw request (the literal body about to hit the API). No membrane or agent-framework changes needed.

Best-effort: failures to append to the log are swallowed so they never break inference.

## Test plan

- [x] Smoke launch: log file created on first inference call
- [x] Format: one JSON line per request, parseable with \`jq\`
- [x] Verified end-to-end during Lena debugging — captured the exact \`temperature: 0\` stowaway that the TUI was hiding

## Sample output

\`\`\`json
{"ts":"2026-05-10T17:40:25.306Z","normalizedConfig":{"model":"claude-opus-4-7","maxTokens":3000},"rawRequest":{...}}
\`\`\`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,7 +19,7 @@ import { Membrane, AnthropicAdapter, NativeFormatter } from '@animalabs/membrane
 import { AgentFramework, AutobiographicalStrategy, PassthroughStrategy, WorkspaceModule, type Module, type MountConfig } from '@animalabs/agent-framework';
 import { resolve, join, basename } from 'node:path';
 import { appendFile, mkdir, stat, rename } from 'node:fs/promises';
-import { appendFileSync, renameSync, readFileSync, existsSync } from 'node:fs';
+import { readFileSync, existsSync } from 'node:fs';
 import { FrontdeskStrategy } from './strategies/frontdesk-strategy.js';
 import { SubagentModule } from './modules/subagent-module.js';
 import { LessonsModule } from './modules/lessons-module.js';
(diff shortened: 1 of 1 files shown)


Anarchid:
# PR #22 Roast — `feat(host): persistent LLM-call log via membrane.beforeRequest hook`

**Author:** antra-tess · **Branch:** `feat/llm-call-log` · **Changed:** `src/index.ts` (+69/-1), `package.json` (1 line) · **Reviewed at:** 2026-05-11

---

## Severity: 🟡 MEH

The feature does what it says on the tin. It correctly piggy-backs on `MembraneConfig.hooks.beforeRequest`, captures both the normalized and raw request, and the bug it was born to catch (the stowaway `temperature: 0`) is exactly the kind of post-mortem this enables. So far, so good. Now sit down. We have things to discuss.

---

## Tooling

```bash
$ npx tsc --noEmit
src/commands.ts(98,16): error TS2322  # pre-existing: dual context-manager copies
src/commands.ts(102,3): error TS2322  # pre-existing
... 30+ pre-existing errors, none in src/index.ts
```

No new type errors. The PR's only touched file (`src/index.ts`) is clean. Tests not run — `bun test` is the registered runner and this PR doesn't touch test files; running the suite blindly to satisfy a checklist would just print pre-existing noise.

---

## Findings

### 🟠 CONCERNING

#### `src/index.ts:621` — Synchronous I/O on every inference, in the hot path

```ts
appendFileSync(llmCallLogPath, JSON.stringify(entry) + '\n');
```

`beforeRequest` fires before every LLM call. You have already imported `appendFile` from `node:fs/promises` ten lines above. You chose `appendFileSync` anyway, then imported a *second* fs module to do it. There is no comment justifying why this needs to block the event loop. A TUI host running multiple agents — which is the documented use case for this project — will stall on disk every time the rotation threshold lines up with an `fsync` flush.

**Fix:** Use `appendFile` (the async one, already imported). Fire-and-forget the promise *inside* the existing try-block — best-effort semantics survive intact:

```ts
appendFile(llmCallLogPath, JSON.stringify(entry) + '\n').catch(() => {});
```

The hook is allowed to return synchronously (`return rawRequest` afterwards is fine); the membrane contract doesn't require the side-effect to be awaited.

#### `src/index.ts:595` — Standard `Number(...) || N` footgun on a numeric env var

```ts
const ROTATE_AT = Number(process.env.LLM_CALL_LOG_ROTATE_AT) || 50;
```

If an operator sets `LLM_CALL_LOG_ROTATE_AT=0` — a reasonable encoding of "disable rotation" or "rotate every entry" depending on whose mental model you adopt — they get 50 anyway, because `||` treats `0` as falsy. Same for `NaN`. Either parse explicitly and validate, or use `??` after a deliberate `parseInt`:

```ts
const parsed = Number.parseInt(process.env.LLM_CALL_LOG_ROTATE_AT ?? '', 10);
const ROTATE_AT = Number.isFinite(parsed) && parsed > 0 ? parsed : 50;
```

#### `package.json:11` — Unrelated change smuggled into a "feat(host)" PR

```diff
-"postinstall": "test -d web && cd web && bun install --production=false && bun run build || true"
+"postinstall": "test -d web && cd web && bun install && bun run build || true"
```

This has nothing to do with the LLM call log. It doesn't appear in the PR body. `--production=false` was forcing devDependencies during the postinstall (so the `web/` build can find `vite`/`tsc`/whatever). Dropping the flag relies on Bun's default behavior including devDeps in `bun install` — which it does *unless* `NODE_ENV=production`, in which case devDeps are silently skipped and the postinstall build fails with cryptic "command not found" errors.

If you tested this in a `NODE_ENV=production` install, say so in the PR. Otherwise, either revert this line or make it its own PR with a one-sentence justification. Tucking it into an unrelated feature commit is how production deploys break on Wednesday afternoon.

### 🟡 MEH

#### `src/index.ts:600`, `624` — Empty catches that swallow operator-relevant failures

```ts
try {
  renameSync(llmCallLogPath, rotated);
} catch {
  // File may not exist yet; ignore.
}
```

The justification ("file may not exist yet") covers `ENOENT`. It does not cover `EPERM`, `EBUSY`, `ENOSPC`, or `EXDEV` (when `dataDir` ends up cross-device, e.g. tmpfs symlink). Same story at line 624 — disk full is "best effort; never break inference," sure, but at minimum a `console.warn` once per session would let the operator notice that their post-mortem log is now a fiction.

**Fix:** Narrow the swallow to `ENOENT`, log everything else once:

```ts
catch (e: any) {
  if (e?.code !== 'ENOENT') console.warn('[llm-call-log] rotate failed:', e?.code);
}
```

#### `src/index.ts:559–573` — `countLines` reads the entire file to count newlines

```ts
const buf = readFileSync(path, 'utf8');
...
for (let i = 0; i < buf.length; i++) if (buf.charCodeAt(i) === 10) n++;
```

At a 50-line cap this is bounded, sure. But you only call it once at startup specifically to decide "do I rotate now?", and the answer you actually want is "is the existing file ≥ 50 lines?" — a question answerable by an `fs.stat` + a streaming line counter that bails at 50. Reading the whole file is unnecessary, and the JSDoc claims "non-empty lines" while the code happily counts a trailing partial line whether it's empty or not. Mismatched docstring, but mostly: this is sync I/O in startup path for no good reason.

Cleaner: track the count in a sidecar file, or just `stat` the size and call it large enough to rotate at ~50KB. Even simpler — don't bother with startup rotation at all. Let the *first* `beforeRequest` of a session check-and-rotate, removing this whole function.

#### `src/index.ts:584–593` — A 10-line comment explaining a 30-line feature

The block comment is longer than the rotation code itself. Half of it ("Useful for post-mortem debugging…", "Old files are kept (no auto-prune)") belongs in the PR description, which already says all this. Comments rot; PR bodies don't get re-read against the code. Trim to one line on the env var and the rotation policy, delete the rest.

#### `src/index.ts:20` — Mixed sync/async fs imports

```ts
import { appendFile, mkdir, stat, rename } from 'node:fs/promises';
import { appendFileSync, renameSync, readFileSync, existsSync } from 'node:fs';
```

The file uses async fs everywhere else for exactly the reasons you'd expect. Adding a parallel sync-fs import the first time someone needs a hot-path log is a smell — once this pattern is in the codebase, the next person will reach for `appendFileSync` because "we already do it." If you fix finding #1, this entire second import line goes away.

### 🟢 NOT A COMPLAINT, BUT WORTH NOTING

The `rawRequest` may contain message content that includes secrets, PII, or user input that wasn't logged anywhere else in the system. The PR's whole *point* is that this is the literal body about to hit the API, so this is by design — but `llm-calls.jsonl` is now the most secret-laden file in `dataDir`. A line in the recipe docs or README that it should be `chmod 600` and excluded from any tarball/upload tooling would not be amiss. Not blocking.

---

## Verdict

The intent is good and the integration point is correct. But this PR landed a sync-I/O hot-path call, a classic `|| N` env-var bug, and a sneaky `package.json` change that has nothing to do with the title. Fix #1 (async fs) and #3 (revert or justify the postinstall change) before merging; the rest can ride along or follow up. The feature itself has already paid for itself once during the Lena debugging session, per the PR body — that's a real result, not a vibe.

PR #23, opened 2026-05-10
fix(commands): branch handlers honor name-keyed switchBranch + async strategy reinit

antra-tess:
## Summary

Pre-fix bugs in \`/undo\`, \`/redo\`, \`/checkpoint\`, \`/restore\`, \`/checkout\` — all surfaced when [anima-research/context-manager#14](https://github.com/anima-research/context-manager/pull/14) (strategy persistence) landed:

1. **id vs name confusion.** \`ContextManager.branchAt\` returns the new branch's *name*, but several handlers stored \`currentBranch.id\` (chronicle's internal numeric id) on the redo / checkpoint stacks. \`switchBranch\` is name-keyed — passing an id raised \`Branch not found\`. **\`/redo\`, \`/restore\`, \`/checkout\` were broken end-to-end.**

2. **No await on async switchBranch.** \`switchBranch\` is now async because it re-initializes the strategy on the new branch (e.g. AutobiographicalStrategy reloads its persisted summaries / pins / mergeQueue from the destination branch's chronicle view). Handlers called it without awaiting, so the strategy could still be on the prior branch's state when the next compile / send-message ran. With autobio: undo + send-message could feed the agent the wrong recall pairs.

## Fix

Uses the existing \`asyncWork\` mechanism on \`CommandResult\` (same pattern as \`/newtopic\`) to wait for switchBranch + strategy reinit before reporting success. The TUI already gates input on \`status === 'thinking'\` while \`asyncWork\` is in flight, so the user can't fire a new turn during the race.

Branch identifiers stored consistently as **names** on undo/redo/checkpoint stacks. Reads fall back to \`branchId\` for legacy state-point entries that might still hold a chronicle id.

## Test plan

- [x] \`bun test\` — 203/205 pass (same 2 pre-existing failures on main: FleetTreeAggregator e2e + CWD-relative allowlist; unrelated)
- [ ] Live verification with autobio: undo, send a different message, verify the response reflects the new branch's strategy state (not the abandoned one)

## Caveats

Without [PR #22 (LLM call log)](https://github.com/anima-research/connectome-host/pull/22), regressions in this area are nearly impossible to debug from operator logs — flagging both are needed for safe ongoing operation against opus-4-7+ and persistent strategies.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -613,28 +613,46 @@ function handleUndo(app: AppContext): CommandResult {
     return { lines: [{ text: 'Nothing to undo (no agent messages found).', style: 'system' }] };
   }
 
+  // Snapshot the current branch's NAME so /redo can return here. Chronicle's
+  // switchBranch is keyed by name, not the internal numeric id.
+  const currentBranch = cm.currentBranch();
+  const newBranchName = `undo-${Date.now()}`;
+
+  let createdBranchName: string;
   try {
-    // Save current state for redo
-    const currentBranch = cm.currentBranch();
-    bs.redoStack.push({
-      branchId: currentBranch.id,
-      branchName: currentBranch.name,
-    });
-
-    // Create a new branch from the undo point
-    const newBranchId = cm.branchAt(undoPoint, `undo-${Date.now()}`);
-    cm.switchBranch(newBranchId);
-
-    bs.undoStack.push({
-      branchId: newBranchId,
-      branchName: `undo-${Date.now()}`,
-      messageId: undoPoint,
-    });
-
-    return { lines: [{ text: `Undone. Switched to branch ${newBranchId}.`, style: 'system' }], branchChanged: true };
+    // branchAt returns the new branch's NAME (the only thing switchBranch accepts).
+    createdBranchName = cm.branchAt(undoPoint, newBranchName);
   } catch (err) {
-    return { lines: [{ text: `Undo failed: ${err}`, style: 'system' }] };
+    return { lines: [{ text: `Undo failed (branchAt): ${err}`, style: 'system' }] };
   }
+
+  bs.redoStack.push({
+    branchId: currentBranch.name, // store NAME — switchBranch expects name
+    branchName: currentBranch.name,
+  });
+  bs.undoStack.push({
+    branchId: createdBranchName,
+    branchName: createdBranchName,
+    messageId: undoPoint,
+  });
+
+  // switchBranch is async — it re-initializes the strategy on the new branch
+  // (autobio reloads its persisted summaries / pins / mergeQueue). Without
+  // awaiting, the next /compile or send-message could see stale strategy
+  // state from the prior branch.
+  const asyncWork = Promise.resolve(cm.switchBranch(createdBranchName))
+    .then(() => ({
+      lines: [{ text: `Undone. Switched to branch "${createdBranchName}".`, style: 'system' as const }],
+      branchChanged: true,
+    }))
+    .catch(err => ({
+      lines: [{ text: `Undo failed (switchBranch): ${err}`, style: 'system' as const }],
+    }));
+
+  return {
+    lines: [{ text: `Undoing → branch "${createdBranchName}"...`, style: 'system' }],
+    asyncWork,
+  };
 }
 
 function handleRedo(app: AppContext): CommandResult {
@@ -647,12 +665,23 @@ function handleRedo(app: AppContext): CommandResult {
   }
 
   const point = bs.redoStack.pop()!;
-  try {
-    cm.switchBranch(point.branchId);
-    return { lines: [{ text: `Redone. Switched to branch ${point.branchName}.`, style: 'system' }], branchChanged: true };
-  } catch (err) {
-    return { lines: [{ text: `Redo failed: ${err}`, style: 'system' }] };
-  }
+  // Use branchName (the canonical key for switchBranch), with branchId as
+  // a fallback for legacy entries that may still hold the chronicle id.
+  const target = point.branchName || point.branchId;
+
+  const asyncWork = Promise.resolve(cm.switchBranch(target))
+    .then(() => ({
+      lines: [{ text: `Redone. Switched to branch "${target}".`, style: 'system' as const }],
+      branchChanged: true,
+    }))
+    .catch(err => ({
+      lines: [{ text: `Redo failed: ${err}`, style: 'system' as const }],
+    }));
+
+  return {
+    lines: [{ text: `Redoing → branch "${target}"...`, style: 'system' }],
+    asyncWork,
+  };
 }
 
 function handleCheckpoint(app: AppContext, name?: string): CommandResult {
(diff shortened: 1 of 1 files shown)


Anarchid:
# PR #23 Roast — `fix(commands): branch handlers honor name-keyed switchBranch + async strategy reinit`

**Author:** antra-tess · **Branch:** `fix/branch-handlers-async-and-name-keyed` · **Changed:** `src/commands.ts` (+88/-37) · **Reviewed at:** 2026-05-11

---

## Severity: 🟠 CONCERNING

Two real bugs identified, two real bugs fixed. `/redo`, `/restore`, and `/checkout` were demonstrably broken before this PR because they passed Chronicle's numeric branch id to a name-keyed `switchBranch`. The autobio strategy staleness during un-awaited switches is also a real footgun. So the *fix* is necessary.

That said, the PR papers over the underlying type-system lie that caused the bug in the first place, and the most important verification step on the test plan is unchecked.

---

## Tooling

```bash
$ npx tsc --noEmit
# 30+ pre-existing errors (dual context-manager copies, MapIterator with target<es2015)
# None in src/commands.ts — the PR's only changed file is clean.

$ bun test
 205 pass
 0 fail
Ran 205 tests across 22 files. [16.16s]
```

PR body claims `203/205 pass (same 2 pre-existing failures)`. The full suite *now* reports 205/205. Either the two failures were flaky and self-resolved, or they're environment-dependent and the PR author's environment had something this one doesn't. Worth a one-line note in the PR — "actually all green here" is information.

---

## Findings

### 🟠 CONCERNING

#### `src/commands.ts:46–50` — `StatePoint.branchId: string` now lies

```ts
export interface StatePoint {
  branchId: string;
  branchName: string;
  messageId?: string;
}
```

After this PR, *every write* into `branchId` stores a name:
- `commands.ts:629` `branchId: currentBranch.name`
- `commands.ts:634` `branchId: createdBranchName`
- `commands.ts:698` `branchId: branch.name`

And every read either prefers `branchName` (`point.branchName || point.branchId`) or treats the value as a name and passes it to `switchBranch`. The `branchId` field is dead data that the next person to touch this code will fill with `branch.id` again because *that is what the field name says to do*. The bug you just fixed will recur.

**Fix:** Either rename the field (`branchKey` or just collapse to `branchName`), or convert the interface to a discriminated shape:

```ts
export interface StatePoint {
  branchName: string;          // the only thing switchBranch will accept
  messageId?: string;
}
```

If you keep `branchId` for backward compat with some persisted state I haven't found, document that explicitly. As-is, the type system is lying to its callers.

#### `src/commands.ts:667, 727, 776` — Fallback against a state shape that can't exist

```ts
const target = point.branchName || point.branchId;
```

Three handlers do this. The comment says "fallback for legacy entries that may still hold a chronicle id." Let's audit that claim. `BranchState` is created in-memory by `createBranchState()` (line 62) and never persisted — there is no disk format, no session save, no migration path. After this PR lands, every push into the stack writes `branchName` to *both* fields (see lines 629, 634, 697–701). The only way `point.branchName` is empty at read time is if someone introduces a new push site that omits it. There is no historical state to defend against.

So this fallback either (a) defends nothing, or (b) is hedging against a serialization story I can't find. Pick one:
- If (a), delete the fallback. `const target = point.branchName;` is honest.
- If (b), add a one-line comment naming the serialization source (`recipes/*.json`? session manager?) so the next reader can verify.

#### PR body test plan — Live verification unchecked

> - [x] `bun test` — 203/205 pass
> - [ ] Live verification with autobio: undo, send a different message, verify the response reflects the new branch's strategy state (not the abandoned one)

The *whole point* of bug #2 (un-awaited switchBranch) is that strategy state can get stale across a branch switch in ways unit tests don't catch. The unit tests pass on `main` too. The unchecked checkbox is the only test that actually exercises the bug the PR claims to fix. Either run it before merge, or change the PR title to acknowledge that the async-await change is theoretical and unverified.

### 🟡 MEH

#### `src/commands.ts:643, 672, 730, 782` — Five near-identical Promise plumbing blocks

```ts
const asyncWork = Promise.resolve(cm.switchBranch(target))
  .then(() => ({
    lines: [{ text: '...', style: 'system' as const }],
    branchChanged: true,
  }))
  .catch(err => ({
    lines: [{ text: '... failed: ${err}', style: 'system' as const }],
  }));

return {
  lines: [{ text: '... → branch ...', style: 'system' }],
  asyncWork,
};
```

This pattern repeats verbatim in `handleUndo`, `handleRedo`, `handleRestore`, `handleCheckout`. A tiny helper would halve the diff:

```ts
function switchBranchAsync(
  cm: ContextManager,
  target: string,
  successMsg: string,
  failurePrefix: string,
  progressMsg: string,
): CommandResult {
  const asyncWork = Promise.resolve(cm.switchBranch(target))
    .then(() => ({ lines: [{ text: successMsg, style: 'system' as const }], branchChanged: true }))
    .catch(err => ({ lines: [{ text: `${failurePrefix}: ${err}`, style: 'system' as const }] }));
  return { lines: [{ text: progressMsg, style: 'system' }], asyncWork };
}
```

Not blocking, but the next person to fix a copy-paste typo in one of the four will miss the other three. (See: this PR, where the four handlers had four near-identical bugs to begin with.)

#### `Promise.resolve(cm.switchBranch(...))` — what *is* the contract?

```ts
const asyncWork = Promise.resolve(cm.switchBranch(createdBranchName)).then(...)
```

If `switchBranch` reliably returns a `Promise<void>` post-strategy-persistence (which the PR body asserts: "switchBranch is now async"), then `Promise.resolve(...)` is redundant noise. If it might be sync or might be async ("now async" being a soft transition), then the wrapping is correct but symptomatic of the migration not landing cleanly. Either:
- Confirm `switchBranch: () => Promise<void>` in the context-manager type, then `await cm.switchBranch(target)`.
- Or document that it's deliberately Promise-or-value to keep options open, with a comment at each site.

The diff bakes in a defensive idiom for what should be a settled contract.

#### `src/commands.ts:618` — Stale progress message after failure

```ts
return {
  lines: [{ text: `Undoing → branch "${createdBranchName}"...`, style: 'system' }],
  asyncWork,
};
```

When `asyncWork` rejects, the TUI shows *both* "Undoing → branch X..." and "Undo failed (switchBranch): ...". One event, two lines. Cosmetic, fixable in follow-up — the TUI's `asyncWork` resolution flow could overwrite the in-progress line. Calling it out so it doesn't go unnoticed.

### 🟢 GOOD CALL

- The split between the `try { branchAt }` (sync, sad-path can short-circuit before pushing to undoStack) and the async `switchBranch` (returns via `asyncWork`) is the right shape. Pushing onto `undoStack` only after `branchAt` succeeds means a failed undo doesn't leave a phantom entry. Solid.
- Citing the upstream PR ([context-manager#14](https://github.com/anima-research/context-manager/pull/14)) that introduced the async contract is exactly the kind of trace the reader needs. More PRs should do this.

---

## Verdict

The fix is necessary, the diagnosis is correct, and the patch will solve the immediate bug. But it leaves the lie at the root — `branchId: string` storing a name — in place, which is how the original bug got written in the first place. Fix the type alongside the call sites, run the autobio live verification before merge, and consider deduplicating the four-times-repeated async block. The dependency on PR #22 for debuggability flagged in the caveats is real; ideally land that first, or both together.

PR #24, opened 2026-05-10
feat(host,tui): strategy config passthrough + status-line memory stats

antra-tess:
## Summary

Two related observability + configurability additions, both downstream of [anima-research/context-manager#15](https://github.com/anima-research/context-manager/pull/15) (autobio audit-gaps).

### 1. Strategy config passthrough (\`src/index.ts\`)

The recipe's \`agent.strategy.{...}\` block now forwards 12 additional fields:

\`\`\`
enforceBudget, maxSpeculativeL1s, positionedRecallPairs, recallHeaderTemplate,
targetChunkTokens, mergeThreshold, summaryTargetTokens,
l1BudgetTokens, l2BudgetTokens, l3BudgetTokens,
toolResultMaxLastN, toolUseInputMaxTokens
\`\`\`

These all landed in context-manager via #15. Without passthrough, recipe authors couldn't actually use them — they'd silently be ignored at instantiation. Specifically blocks Lena's recipe, which relies on \`enforceBudget: false\` (no truncation, surface API-side overage instead) and \`maxSpeculativeL1s: 36\` (cap for the new background drain).

### 2. Status-line memory stats (\`src/tui.ts\`)

The TUI status line now shows live memory state when the active strategy implements \`getRenderStats()\`:

\`\`\`
12.3kin 2.1kout 8.5khit 200write  C-v:terse  mem: L1=8/16kt L2=1/2kt tail=770msg/495kt (4 pending, 1 merge)
\`\`\`

- **Per-level**: count + total tokens of visible (un-merged) summaries
- **tail**: message count + token estimate of the recent uncompressed window
- **pending**: chunks queued for compression / merges queued
- **cache**: \`hit\` = read tokens, \`write\` = creation tokens (now populated from \`inference:usage\` events; previously the \`TokenUsage\` fields existed but weren't wired)

Falls back to the older \`getStats()\`-only display for strategies that don't expose \`getRenderStats\` yet.

## Test plan

- [x] Verified live with Lena: status line updates per turn, shows L1 count growing, tail size shrinking as chunks form, eventually shows L2 after merge fires
- [x] Cache fields populate after first inference (`inference:usage` event handler now reads \`cacheRead\` / \`cacheCreation\`)
- [x] No regressions in existing tests (203/205 — same 2 pre-existing failures: FleetTreeAggregator e2e + CWD-relative allowlist)

## Companion PRs

Depends on [anima-research/context-manager#15](https://github.com/anima-research/context-manager/pull/15) for \`getRenderStats()\` + the new strategy config fields. Merge after that lands and a new context-manager publishes.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -299,13 +299,32 @@ async function createFramework(membrane: Membrane, storePath: string, recipe: Re
   // -- Build strategy --
   const strategyConfig = recipe.agent.strategy;
   const strategyType = strategyConfig?.type ?? 'autobiographical';
-  const autobiographicalOpts = {
+  const autobiographicalOpts: Record<string, unknown> = {
     headWindowTokens: strategyConfig?.headWindowTokens ?? 4000,
     recentWindowTokens: strategyConfig?.recentWindowTokens ?? 30000,
     compressionModel: strategyConfig?.compressionModel ?? model,
     autoTickOnNewMessage: true,
     maxMessageTokens: strategyConfig?.maxMessageTokens ?? 10000,
   };
+  // Pass-through additional fields if the recipe sets them (kept narrow:
+  // only fields the recipe schema actually accepts get forwarded).
+  for (const key of [
+    'enforceBudget',
+    'maxSpeculativeL1s',
+    'positionedRecallPairs',
+    'recallHeaderTemplate',
+    'targetChunkTokens',
+    'mergeThreshold',
+    'summaryTargetTokens',
+    'l1BudgetTokens',
+    'l2BudgetTokens',
+    'l3BudgetTokens',
+    'toolResultMaxLastN',
+    'toolUseInputMaxTokens',
+  ] as const) {
+    const v = (strategyConfig as Record<string, unknown> | undefined)?.[key];
+    if (v !== undefined) autobiographicalOpts[key] = v;
+  }
   const strategy = strategyType === 'passthrough'
     ? new PassthroughStrategy()
     : strategyType === 'frontdesk'
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,7 +23,7 @@ import { FrontdeskStrategy } from './strategies/frontdesk-strategy.js';
 import { SubagentModule } from './modules/subagent-module.js';
 import { LessonsModule } from './modules/lessons-module.js';
 import { RetrievalModule } from './modules/retrieval-module.js';
-import type { RecipeWorkspaceMount } from './recipe.js';
+import type { RecipeWorkspaceMount, RecipeStrategy } from './recipe.js';
 import { TuiModule } from './modules/tui-module.js';
 import { TimeModule } from './modules/time-module.js';
 import { FleetModule, type FleetModuleConfig } from './modules/fleet-module.js';
@@ -297,6 +297,14 @@ async function createFramework(membrane: Membrane, storePath: string, recipe: Re
   // No server augmentation needed — gate is wired via FrameworkConfig.gate
 
   // -- Build strategy --
+  //
+  // Build the options object with typed property access — no
+  // `Record<string, unknown>` cast on `strategyConfig`. Every field we
+  // forward is declared on `RecipeStrategy` (see recipe.ts); a typo in a
+  // recipe (e.g. `l1BudgetTokes`) now fails at recipe validation rather
+  // than silently being a no-op at strategy construction. AutobiographicalStrategy
+  // and FrontdeskStrategy share this option bag today; if strategy-specific
+  // fields are ever added, this should split into per-strategy types.
   const strategyConfig = recipe.agent.strategy;
   const strategyType = strategyConfig?.type ?? 'autobiographical';
   const autobiographicalOpts: Record<string, unknown> = {
@@ -306,9 +314,10 @@ async function createFramework(membrane: Membrane, storePath: string, recipe: Re
     autoTickOnNewMessage: true,
     maxMessageTokens: strategyConfig?.maxMessageTokens ?? 10000,
   };
-  // Pass-through additional fields if the recipe sets them (kept narrow:
-  // only fields the recipe schema actually accepts get forwarded).
-  for (const key of [
+  // Forward optional tuning fields when set. The key list is typed
+  // against `RecipeStrategy`, so an unknown field name is a compile
+  // error here rather than a silent no-op at runtime.
+  const passthroughKeys: ReadonlyArray<keyof RecipeStrategy> = [
     'enforceBudget',
     'maxSpeculativeL1s',
     'positionedRecallPairs',
@@ -321,8 +330,9 @@ async function createFramework(membrane: Membrane, storePath: string, recipe: Re
     'l3BudgetTokens',
     'toolResultMaxLastN',
     'toolUseInputMaxTokens',
-  ] as const) {
-    const v = (strategyConfig as Record<string, unknown> | undefined)?.[key];
+  ];
+  for (const key of passthroughKeys) {
+    const v = strategyConfig?.[key];
     if (v !== undefined) autobiographicalOpts[key] = v;
   }
   const strategy = strategyType === 'passthrough'
--- a/src/recipe.ts
+++ b/src/recipe.ts
@@ -20,10 +20,30 @@ import { dirname, isAbsolute, resolve } from 'node:path';
 
 export interface RecipeStrategy {
   type: 'autobiographical' | 'passthrough' | 'frontdesk';
+  // Core window/compression sizing
   headWindowTokens?: number;
   recentWindowTokens?: number;
   compressionModel?: string;
   maxMessageTokens?: number;
+  // Compression/merge tuning passed through to the underlying
+  // autobiographical strategy (and frontdesk, which extends it).
+  // Declared here so a typo in a recipe (e.g. `l1BudgetTokes`) fails
+  // at recipe validation rather than silently defaulting at strategy
+  // construction. Keep in sync with `AutobiographicalConfig` in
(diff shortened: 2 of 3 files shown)


Anarchid:
# PR #24 Roast — `feat(host,tui): strategy config passthrough + status-line memory stats`

**Author:** antra-tess · **Branch:** `feat/tui-mem-stats-and-strategy-passthrough` · **Changed:** `src/index.ts` (+20/-1), `src/tui.ts` (+91/-3) · **Reviewed at:** 2026-05-11

---

## Severity: 🟠 CONCERNING

The status-line memory stats are a genuinely useful piece of observability — the author verified live with Lena that it does what's claimed, and the cache-field wiring fixes a real gap. The strategy passthrough is also necessary plumbing for context-manager#15.

But the passthrough's own comment lies about its invariant, and the type system is deliberately erased at precisely the point where the validation should kick in. This makes a recipe-author footgun *worse* than it was, not better. Two fairly small fixes turn this from concerning into shippable.

---

## Tooling

```bash
$ npx tsc --noEmit
# 30+ pre-existing errors, none in src/index.ts or src/tui.ts.

$ bun test
 205 pass
 0 fail
Ran 205 tests across 22 files. [16.14s]
```

PR body claims `203/205 (same 2 pre-existing failures)`. Same observation as PR #23: full suite is green here. Either the failures are environment-specific or they've self-resolved. Worth a one-line update.

---

## Findings

### 🟠 CONCERNING

#### `src/index.ts:309–327` — The "kept narrow" comment is a lie

```ts
// Pass-through additional fields if the recipe sets them (kept narrow:
// only fields the recipe schema actually accepts get forwarded).
for (const key of [
  'enforceBudget',
  'maxSpeculativeL1s',
  // ... 10 more ...
] as const) {
  const v = (strategyConfig as Record<string, unknown> | undefined)?.[key];
  if (v !== undefined) autobiographicalOpts[key] = v;
}
```

Now look at the schema this comment claims to honor:

```ts
// src/recipe.ts:21–27
export interface RecipeStrategy {
  type: 'autobiographical' | 'passthrough' | 'frontdesk';
  headWindowTokens?: number;
  recentWindowTokens?: number;
  compressionModel?: string;
  maxMessageTokens?: number;
}
```

`RecipeStrategy` does not declare *any* of the 12 fields you just forwarded. The comment asserts a constraint the code actively defeats by casting `strategyConfig` to `Record<string, unknown>`. A recipe author writes `l1BudgetTokens: 4000` → TypeScript says nothing (the cast erases it) → `validateRecipe` says nothing (no unknown-key check at recipe.ts:546–558) → autobio strategy *does* accept it → silent success. Now the recipe author writes `l1BudgetTokes: 4000` (typo) → TypeScript says nothing → validation says nothing → autobio gets default → silent failure.

This is exactly the class of bug PR #22 (the LLM-call log) exists to debug. Add the fields to the schema *first*, then forward them with a typed property access:

```ts
// recipe.ts
export interface RecipeStrategy {
  type: 'autobiographical' | 'passthrough' | 'frontdesk';
  headWindowTokens?: number;
  recentWindowTokens?: number;
  compressionModel?: string;
  maxMessageTokens?: number;
  enforceBudget?: boolean;
  maxSpeculativeL1s?: number;
  positionedRecallPairs?: number;
  recallHeaderTemplate?: string;
  targetChunkTokens?: number;
  mergeThreshold?: number;
  summaryTargetTokens?: number;
  l1BudgetTokens?: number;
  l2BudgetTokens?: number;
  l3BudgetTokens?: number;
  toolResultMaxLastN?: number;
  toolUseInputMaxTokens?: number;
}
```

Then `index.ts` becomes:

```ts
const autobiographicalOpts: AutobiographicalStrategyOptions = {
  headWindowTokens: strategyConfig?.headWindowTokens ?? 4000,
  // ... existing defaults ...
  ...(strategyConfig?.enforceBudget !== undefined && { enforceBudget: strategyConfig.enforceBudget }),
  // ... or use a deliberate spread with explicit type ...
};
```

The comment then *matches the code*. As a bonus, the `Record<string, unknown>` line disappears and `autobiographicalOpts` gets a real type back.

#### `src/index.ts:302` — `autobiographicalOpts: Record<string, unknown>` erases strategy contract

```ts
const autobiographicalOpts: Record<string, unknown> = { ... };
```

Three different strategies (`AutobiographicalStrategy`, `FrontdeskStrategy`, `PassthroughStrategy`) consume this object. Two of them at lines 331–332:

```ts
: strategyType === 'frontdesk'
  ? new FrontdeskStrategy(autobiographicalOpts)
  : new AutobiographicalStrategy(autobiographicalOpts);
```

Do `FrontdeskStrategy` and `AutobiographicalStrategy` accept the same 16 options? If `mergeThreshold` is autobio-specific, what does Frontdesk do when it gets it — silently ignore? Throw? You've made it impossible for TypeScript to answer. Tighten the type to the strategy's actual option interface (export it from agent-framework if not already exposed). This is the second time in this PR the type system is being asked to take a nap.

#### `src/tui.ts:2019–2020` — User-facing status-line string change is unannounced

```ts
-if (tokens.cacheRead > 0) s += ` ${fmtTokens(tokens.cacheRead)}cache`;
+if (tokens.cacheRead > 0) s += ` ${fmtTokens(tokens.cacheRead)}hit`;
+if (tokens.cacheWrite > 0) s += ` ${fmtTokens(tokens.cacheWrite)}write`;
```

The PR body advertises the *cache values being populated* but doesn't mention that the label changed from `cache` to `hit`/`write`. If any operator or screenshot tool greps the status line for the literal string `cache`, they break silently. Probably no one is doing that — but the title says "memory stats," and a string format change in the same line deserves a sentence in the PR description. Cheap to fix: one bullet under "Test plan" → "Status-line format changed: previous `Xcache` is now `Xhit Ywrite`."

### 🟡 MEH

#### `src/tui.ts:350–356` — `as any` in `getRootCM`

```ts
function getRootCM(): { getRenderStats?: () => unknown; getStrategy?: () => { getStats?: () => unknown } } | null {
  try {
    const ag = app.framework.getAgent(rootAgentName);
    if (!ag) return null;
    return ag.getContextManager() as any;
  } catch {
    return null;
  }
}
```

The function declares a perfectly serviceable structural return type. Then casts to `any` instead of casting to that type. The structural type is what tells the rest of the file what's safe to call — the `any` cast throws that away at the boundary. Use:

```ts
return ag.getContextManager() as unknown as { getRenderStats?: ... } | null;
```

or — better — narrow `ContextManager` to publicly expose `getRenderStats`/`getStrategy` upstream once context-manager#15 publishes, so this whole cast goes away. Why is `try`/`catch` needed at all? `getAgent` and `getContextManager` aren't supposed to throw on a known agent name — if they can, the contract is busted and the silent return null hides it.

#### `src/tui.ts:1205–1210` — Output accumulated, others overwritten, no comment

```ts
if (roundUsage.input !== undefined) state.tokens.input = roundUsage.input;
if (roundUsage.output !== undefined) state.tokens.output = (state.tokens.output ?? 0) + (roundUsage.output ?? 0);
if (roundUsage.cacheRead !== undefined) state.tokens.cacheRead = roundUsage.cacheRead;
if (roundUsage.cacheCreation !== undefined) state.tokens.cacheWrite = roundUsage.cacheCreation;
```

`input`, `cacheRead`, `cacheWrite` are *assigned*. `output` is *accumulated*. The reason (input is the running context size for the current turn — latest wins — whereas output is generated tokens that compound across yielding-stream rounds) is correct, but invisible to a reader who didn't write it. One line of comment averts a wrong "consistency fix" PR six months from now:

```ts
// output accumulates across yielding-stream rounds; input/cache are running
// snapshots so the most recent round wins.
```

#### `src/tui.ts:2032–2092` — `formatMemStats` carries two complete rendering paths

The function does the new `getRenderStats()` rendering, falls back to `getStats()` rendering with different field names (`pendingMerges` vs `pending.merges`, `chunksTotal - chunksCompressed` vs `pending.chunks`, etc.). If the plan is for context-manager#15 to land everywhere and the fallback to die, mark it `@deprecated` so we can grep it out later. If both will coexist long-term, fine — but at 60 lines the function is asking to be split (`formatMemStatsRich` / `formatMemStatsLegacy`) so the next reader doesn't have to mentally parse "which case am I in?" on every fix.

#### Empty `catch {}` in `formatMemStats` and `getRootCM`

```ts
} catch {
  return '';
}
```

A status-line formatter throwing means something interesting just broke in the strategy. Returning '' hides it for the user; a one-line `console.warn` (debounced if needed) would let the operator notice. Not blocking — TUIs eat exceptions all the time — but a `// TODO: surface to debug log` comment costs nothing.

### 🟢 GOOD CALL

- Adding `cacheRead`/`cacheWrite` rendering to the status line was overdue. Per the PR body, the `TokenUsage` fields already existed; this PR is wiring them. Closing the loop on a feature that was 90% there is real work.
- The fallback to `getStats()` when `getRenderStats` isn't available is the right shape for landing a host PR before the dependency publishes. The reviewer can verify by reading `formatMemStats` top-down — no surprise behavior.
- Naming the verification done with Lena gives the reader something concrete to imagine. More PRs should do this.

---

## Verdict

The status-line work is fine to ship as-is — minor cleanups, no structural issues. The strategy-config passthrough is half-finished: declaring the fields in `RecipeStrategy` is one type addition that turns "silent footgun" into "compiler error on typo," and the comment that already promises this constraint can stop being aspirational. Until that's done, this PR makes the recipe ergonomics meaningfully worse than before, because nothing else validates `agent.strategy.*` keys.

Recommended sequence: add the 12 fields to `RecipeStrategy`, drop the `Record<string, unknown>` cast on `autobiographicalOpts`, mention the status-line string format change in the body, then merge.

PR #27, opened 2026-05-18
fix(warmup,host): identity handoff from import → warmup → live agent

Anarchid:
## Summary

Fixes the warmup-output-invisible-to-framework bug observed in the wild
on a 4010-message claude.ai-imported session: the autobiography Sonnet
4.5 spent ~16 hours producing was written to `default/autobio:summaries`
while the live agent reads from `agents/${name}/autobio:summaries`, so
the reporter's Claude opened the resumed conversation with no memory of
months of context and reported a December cliff (consistent with the
head window's content + an absent middle).

Two structural bugs plus three pieces of defense-in-depth around them:

- **`warmup-session.ts` namespace fix** (the actual bug):
  `ContextManager.open` was called without a namespace, falling back to
  the strategy's `'default'` sentinel. Live framework uses
  `'agents/${config.name}'`. The two slots never met. Now warmup
  passes `namespace: 'agents/${agentName}'`.
- **Live-session Membrane `assistantParticipant`**: conhost's
  Membrane previously had no `assistantParticipant` set. Compression
  requests don't set it on the request either, so they fell back to
  `'Claude'`. For any non-`Claude` agent name (current conhost
  default is `'agent'`), every stored assistant turn got mapped to
  `role: 'user'` and tool_use blocks ended up in user messages,
  which the API rejects. Symptom in the reporter's `llm-calls.jsonl`:
  13 byte-identical 56136-byte single-message-`role:user`
  compression attempts. Now the Membrane is configured with
  `assistantParticipant: agentName`.
- **Sidecar `agentName` metafile**: the importer records the agent name
  it used into `<id>.import-source.json`. `SessionManager.getImportSource()`
  surfaces it. Warmup reads it back so the operator doesn't have to
  pass `--agent` twice in a row to keep three places in sync.
- **Defaults flip to `"Claude"`**: importer + warmup defaults change
  from `'agent'` to `'Claude'` for the claudeai-revival workflow.
  Other recipes are unaffected (their stored data already uses
  whatever name they chose).
- **Revival recipe pins `agent.name: "Claude"`** to match the new
  default end-to-end.

## Commit-by-commit

| # | SHA | What |
|---|---|---|
| C1 | `0127237` | importer writes `agentName` to sidecar; `SessionManager.getImportSource()` + `ImportSource` type |
| C2 | `09b4cb5` | defaults flip to `"Claude"`; warmup `--agent → sidecar → "Claude"` priority; docs |
| C3 | `5ba4ae4` | warmup `namespace: 'agents/${agentName}'` — **the fix** |
| C4 | `c539159` | live-session Membrane `assistantParticipant: agentName` |
| C5 | `e1507a6` | `recipes/claude-export-revive.json` `agent.name: "Claude"` |
| C6 | `f982a0c` | 6 tests pinning the handoff contracts (no LLM in CI) |

## Test plan

- [ ] `bun test test/warmup-handoff.test.ts test/claudeai-export-importer.test.ts test/autobio-progress-snapshot.test.ts` — local: 40/40 pass
- [ ] Fresh import of a small claude.ai conversation; verify the sidecar contains `"agentName": "Claude"`
- [ ] Warmup that session; verify `bun scripts/check-autobio-namespace.ts <session>` (or direct `listStates` probe) shows summaries at `agents/Claude/autobio:summaries` and nothing at `default/autobio:summaries`
- [ ] Open the session in conhost against `recipes/claude-export-revive.json`; verify the agent's compiled prompt includes summaries (not just head + tail) and that subsequent post-warmup compressions land at the same slot

## Out of scope, deliberately

- The `token_budget` / `flag` block-type skip in the importer (stashed locally; debated — keeping these as `[unknown_block]` placeholders may better preserve the original model's read of the conversation; deserves its own design call before shipping).
- Pushing the runtime tool_result-split into the Membrane formatter (the architecturally-cleaner Bug 4/6/8 fix). The defense-in-depth versions are already in `context-manager` `fix/append-with-assigned-id`; the formatter shift can ride later as a separate cleanup PR once that lands.
- Migration script for users with existing `default/*` summaries. Per discussion, fresh re-import is preferred over data-stitching so the full pipeline gets validated end-to-end.

## Depends on

`f05baa6 chore(deps): bump chronicle to ^0.2.0, context-manager to ^0.5.2` (already at the base of this branch; can be folded if it merges first).

### DIFF
--- a/docs/claudeai-evacuation.md
+++ b/docs/claudeai-evacuation.md
@@ -85,7 +85,7 @@ Useful flags:
 | Flag | Purpose |
 |---|---|
 | `--out <dir>` | Conhost data dir (default `./data`) |
-| `--agent <name>` | Participant name for assistant turns (default `agent` — **must equal the recipe's `agent.name`** or the API request builder will assign the wrong role) |
+| `--agent <name>` | Participant name for assistant turns (default `Claude`). The chosen value is recorded in the import-source sidecar; warmup and the bundled `claude-export-revive.json` recipe pick it up automatically. Override only if you have a reason to depart from `Claude`. |
 | `--filter <regex>` | Case-insensitive name regex; combines with the interactive picker |
 | `--dry-run` | Parse + report, don't write |
 | `--no-interactive` | Skip the picker; import everything (after `--filter`) |
@@ -148,7 +148,7 @@ Useful warmup flags:
 |---|---|
 | `--data-dir <dir>` | Conhost data dir (default `./data`) |
 | `--model <id>` | Compression model (default `claude-sonnet-4-5-20250929`) |
-| `--agent <name>` | Participant name for assistant turns (default `agent`). **Must match the value used at import** — otherwise Membrane will map assistant messages to role `user` and the API will reject the compression request. |
+| `--agent <name>` | Participant name for assistant turns. If omitted, read from the session's import-source sidecar; falls back to `Claude` otherwise. Must equal the value the session was imported with — Membrane formats the assistant role by string-comparing message participants against this name. |
 | `--max-spend <usd>` | Soft cap — halts gracefully when running cost hits the cap. Re-run to resume. |
 | `--l1-budget <n>`, `--l2-budget <n>`, `--l3-budget <n>` | Autobio tier token budgets |
 | `--merge-threshold <n>` | L1→L2 / L2→L3 merge threshold (default 6) |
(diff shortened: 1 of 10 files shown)


Anarchid:
QA pass addressed in 4 follow-up commits, all in-branch:

| # | SHA | Finding |
|---|---|---|
| C7 | `80d0c7c` | 🟠 Finding 2 — extract `resolveAgentName({explicit, sidecar, default})` returning `{name, source, mismatch?}`; 8 tests pin the priority chain (incl. empty-string fall-through and mismatch detection) |
| C8 | `2f6a93b` | 🟡 Findings 5, 6 — warmup adopts the helper; nested ternary in log → uses `source` directly; trims the 9-line block above `ContextManager.open` to two lines |
| C9 | `44b39e8` | 🟠 Finding 1 + 🟡 Finding 4 + 🟡 Finding 5 — conhost now consults `sidecar.agentName` when `recipe.agent.name` is absent; `console.warn` on recipe-vs-sidecar mismatch (parallels warmup's new warning); the three duplicate `recipe.agent.name \|\| 'agent'` sites in src/index.ts unified through `AppContext.agentName` plumbed from main(); trims the 8-line block above the Membrane construction |
| C10 | `32677d4` | 🟡 Finding 3 — `getImportSource` now gates on `typeof === 'object' && !== null && !Array.isArray` before the `as ImportSource` cast; 4 tests cover `null`/array/string/number JSON roots |

The decision on Finding 1 was "consult sidecar in conhost when recipe.agent.name is absent" rather than "flip the conhost default to Claude" — the latter would silently change `agent.name` for existing native sessions that have been running under `agent`. The sidecar-first approach scopes the new behavior to imported sessions where the sidecar exists.

`switchSession` keeps the startup-resolved agentName rather than re-resolving per session; documented inline. Per-session re-resolution would matter only for the fringe case of "custom recipe with no agent.name, imports created under different --agent values, user switches between them" — not the canonical revival flow, and would also need to rebuild the Membrane mid-run, which is a bigger refactor than this PR's scope.

Test totals: 49/49 pass across `agent-name`, `warmup-handoff`, `claudeai-export-importer`, and `autobio-progress-snapshot`. Type check clean.

Out of scope from the QA pass, deliberately:
- The chronicle ^0.2.0 / npm-publication timing issue — separate concern as flagged.
- Pre-existing fleet/headless/subagent failures — verified to fail on main too.

PR #28, opened 2026-05-20
fix(webui): route cancel-subagent to owning fleet child

Anarchid:
## Summary

- The admin WUI's "stop" button on a subagent leaf failed with `subagent module not loaded` whenever the subagent lived inside a fleet child (e.g. a Clerk-spawned fork). The conductor has no `SubagentModule`, so the cancel hit the wrong process.
- Extend `cancel-subagent` with an optional `childName`. When set, the conductor forwards a new `cancel-subagent` IPC command to that fleet child's headless runtime, which calls its local `SubagentModule.cancelSubagent()`. The child replies with `cancel-subagent-result` (corrId-correlated), and the conductor translates it into a normal `command-result` line for the originating browser. Without `childName` the conductor-local path is unchanged.
- The SPA reads the owning fleet child name from the subagent leaf's `child-event-agent` stream source and threads it through `onCancelSubagent`. `stopFocusedNode` cancels the subagent specifically for a subagent inside a fleet child, rather than killing the whole child as a workaround.

## Why

Surfaced while investigating a zombie subagent that survived the parent's `sync-fork` timeout and stayed registered in the Clerk's `framework.agents` for ~30h, eventually double-replying to a user message alongside the main Clerk. The conductor's "stop" button was the only containment verb available, and it didn't reach the child where the subagent actually lives — operators had to restart the entire fleet child to clear one zombie.

The underlying ephemeral-agent leak (sync-fork timeout not cancelling inner inference; cleanup listener missing on natural `end_turn`) is being addressed separately in `agent-framework`. This PR is the containment-side fix: even when a zombie escapes that race, an operator can now stop just the zombie without nuking the host child.

## Test plan

- [ ] `npx tsc --noEmit` clean in both `connectome-host/` and `connectome-host/web/` (verified locally)
- [ ] Manual WUI test: open admin SPA, locate a running subagent under a fleet child, press stop → reply line reads `cancelled subagent <name>` (or `subagent <name> not cancelled (subagent not running)` if already gone), and the subagent disappears from the tree
- [ ] Manual WUI test: same flow for a local (conductor-owned) subagent still works — the path without `childName` is unchanged
- [ ] Manual WUI test: stop button on a fleet child itself (not a subagent inside it) still calls `fleet-stop` as before

### DIFF
--- a/src/headless.ts
+++ b/src/headless.ts
@@ -329,6 +329,30 @@ export async function runHeadless(app: AppContext, argv: string[] = []): Promise
         }
         return;
       }
+      case 'cancel-subagent': {
+        const mod = app.framework.getAllModules().find((m) => m.name === 'subagent') as
+          | { cancelSubagent(name: string): boolean }
+          | undefined;
+        if (!mod) {
+          emit({
+            type: 'cancel-subagent-result',
+            corrId: cmd.corrId,
+            name: cmd.name,
+            cancelled: false,
+            reason: 'subagent module not loaded',
+          });
+          return;
+        }
+        const ok = mod.cancelSubagent(cmd.name);
+        emit({
+          type: 'cancel-subagent-result',
+          corrId: cmd.corrId,
+          name: cmd.name,
+          cancelled: ok,
+          ...(ok ? {} : { reason: 'subagent not running' }),
+        });
+        return;
+      }
       default: {
         const t = (cmd as { type?: unknown }).type;
         log(`unknown command type: ${String(t)}`);
--- a/src/modules/fleet-module.ts
+++ b/src/modules/fleet-module.ts
@@ -1597,6 +1597,18 @@ export class FleetModule implements Module {
     catch { return false; }
   }
 
+  /** Cancel a specific in-process subagent inside a fleet child. Response
+   *  arrives as a `cancel-subagent-result` event on the child's event stream;
+   *  the WUI handler matches on corrId to route the result back to the
+   *  originating browser client. Returns false if the named child isn't
+   *  running. */
+  cancelSubagentOnChild(childName: string, name: string, corrId?: string): boolean {
+    const child = this.children.get(childName);
+    if (!child || !child.socket) return false;
+    try { this.sendToChild(child, { type: 'cancel-subagent', name, corrId }); return true; }
+    catch { return false; }
+  }
+
   private async killChild(child: FleetChild): Promise<void> {
     if (child.status === 'exited' || child.status === 'crashed') return;
     const proc = child.process;
--- a/src/modules/fleet-types.ts
+++ b/src/modules/fleet-types.ts
@@ -33,7 +33,12 @@ export type IncomingCommand =
   /** Pull a recursive listing of one mount in the child. Response is `workspace-tree-snapshot`. */
   | { type: 'request-workspace-tree'; mount: string; corrId?: string }
   /** Read a workspace file from the child. Response is `workspace-file-snapshot`. */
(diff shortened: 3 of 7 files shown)


Anarchid:
## :robot: Insufferable Code Review — Severity :green_circle: ACCEPTABLE (with :yellow_circle: paper cuts)

Tooling clean, tests pass (292/292), motivation unambiguous, wire-path coherent. Five trims worth making before this lands.

### :yellow_circle: 1. The trust-boundary test grew nothing
`src/web/protocol.ts:498` extended `cancel-subagent` to accept an optional `childName`. `test/web-ui-protocol.test.ts` opens with a mission statement calling itself *"the trust boundary between unauthenticated wire bytes and handlers that … spawn processes, or fan out to fleet children"* — and got zero new cases. Six lines: `v.childName = 123` rejected, `v.childName = ""` rejected, `v.childName = undefined` accepted, `v.childName = "clerk-1"` accepted. Add them.

### :yellow_circle: 2. No round-trip test for the new corrId path
`test/fleet-route.test.ts`, `test/fleet-commands.test.ts`, and `test/mock-headless-child.ts` already assert that `request-lessons` / `request-workspace-*` round-trip across the corrId map. The new `cancel-subagent` IPC command + `cancel-subagent-result` event + `routeFleetRequest('cancel-subagent', …)` dispatcher are operationally load-bearing — they're the verb of last resort to clear a zombie that escaped the sync-fork race. Pattern is sitting right there in the harness.

### :yellow_circle: 3. `getAllModules().find('subagent') as { cancelSubagent(...) }` — now three copies
- `src/modules/web-ui-module.ts:825` (pre-existing, conductor-local path)
- `src/headless.ts:330` (new, child path)
- plus the analogous local cast pattern

Three is the count at which you factor a helper. The day someone changes `SubagentModule.cancelSubagent`'s signature, all three sites silently keep compiling because every one of them launders the lookup through `as`. Not a blocker, but you're painting the same wall a third time.

### :yellow_circle: 4. Two wire shapes for the same failure
When `SubagentModule` is absent:
- Local path (`web-ui-module.ts:830-832`): `{ type: 'error', message: 'subagent module not loaded' }`
- New routed path (`headless.ts:333-340` → `web-ui-module.ts:492-505`): `{ type: 'command-result', lines: [{ text: 'subagent X not cancelled (subagent module not loaded)', style: 'tool' }] }`

Same condition, two SPA renderings. Pick one.

### :yellow_circle: 5. Cosmetic asymmetry between Tree.tsx and App.tsx
- `web/src/Tree.tsx:82` — `props.onCancelSubagent(props.node.label, owningFleetChild())`
- `web/src/App.tsx:357` — `cancelSubagent(src.agentName, src.childName)`

Both equal `node.name` today (per `tree.ts:201-207`), so not a defect — but if someone ever changes `label` to a display-friendly variant (`"clerk-1 (idle)"`), `Tree.tsx` will silently start cancelling the wrong agent. Read from `streamSource.agentName` in both places.

---

## Tooling

```
$ npx tsc --noEmit             # backend
(clean, no output)

$ cd web && npx tsc --noEmit    # SPA
(clean, no output)

$ bun test
292 pass, 0 fail, 799 expect() calls — 28 files, 17.6s
```

## Verdict

Real fix to a real containment gap. Clean diff, honest commit message that names the zombie incident, type-checks both halves, breaks nothing. The architectural choice — extending the existing pending-corrId routing rather than inventing a parallel mechanism — is the right one. **Ship it after #1 and #2** — operators stopping a zombie is exactly the verb you don't want to learn is broken by way of a second incident.

— review generated by `/roast`

PR #32, opened 2026-05-21
feat(tui): dedicated render lane for thinking blocks

Anarchid:
## Summary

Extended-thinking content was streaming into the same WHITE text element as the assistant's reply, indistinguishable from it. Once an assistant turn finished, only the visible-text part lived on as a `TextRenderable` — thinking blocks loaded from Chronicle on session restart were dropped entirely by `loadSessionHistory`.

This PR adds a proper lane for thinking content in the TUI:

- New `THINKING_DIM` colour (\`#8a7aa8\`, muted violet).
- Stream rendering tracks \`currentStreamBlockType\` and routes tokens to a separately-styled \`TextRenderable\` per block lane. Switching is driven by the new \`inference:content_block\` trace event from agent-framework on \`block_start\`; same-lane re-entry is a no-op.
- Thinking elements get a \`💭 \` prefix so they read clearly even without colour.
- \`inference:tokens\` carries a belt-and-braces lane correction: if a chunk's \`blockType\` disagrees with what \`block_start\` announced (or block_start was missed), the lane switches before appending.
- \`loadSessionHistory\` now renders \`block.type === 'thinking'\` past blocks (dim violet, \`💭 \` prefix) instead of silently dropping them.

## Dependencies

- \`@animalabs/agent-framework\`: anima-research/agent-framework#35 — emits the \`inference:content_block\` trace events and the \`blockType\` tag on \`inference:tokens\`.
- \`@animalabs/membrane\`: antra-tess/membrane#19 — fixes the native yielding path so block events actually fire when thinking is on.

This PR is correct on its own (the trace handler is purely additive and a missing event is harmless) but only delivers user-visible behaviour once the two upstream PRs land and the deps bump.

## Test plan

- [x] \`bun test\` — 292/292 pass
- [x] \`npx tsc --noEmit\` — clean
- [x] Manual end-to-end against \`data-evac\`: continued a session with \`recipe.agent.thinking.enabled = true\` against Opus, thinking now streams in dim violet with the \`💭 \` prefix; historical thinking blocks (where Chronicle stored them structurally) replay correctly via \`loadSessionHistory\`.

## Not in scope

- Subagent peek and fleet-child views — they have their own line-merging buffers; can layer the same distinction in later if needed.
- WebUI rendering — the wire protocol would need a new variant carrying \`blockType\`. Separate PR.

### DIFF
--- a/src/tui.ts
+++ b/src/tui.ts
@@ -128,6 +128,7 @@ const RED = '#cc0000';
 const GRAY = '#888888';
 const DIM_GRAY = '#555555';
 const WHITE = '#cccccc';
+const THINKING_DIM = '#8a7aa8';
 
 // ---------------------------------------------------------------------------
 // Main
@@ -355,8 +356,11 @@ export async function runTui(app: AppContext): Promise<void> {
     }
   }
 
+  let currentStreamBlockType: 'text' | 'thinking' | 'tool_call' | 'tool_result' = 'text';
+
   function beginStream() {
     currentStreamBuffer = '';
+    currentStreamBlockType = 'text';
     currentStreamText = new TextRenderable(renderer, {
       id: `stream-${++messageCounter}`,
       content: '',
@@ -364,6 +368,38 @@ export async function runTui(app: AppContext): Promise<void> {
     });
     scrollBox.add(currentStreamText);
     streaming = true;
+  }
+
+  /**
+   * Switch the active stream element to a different block lane. Called from
+   * inference:content_block (block_start). If the new block is the same lane
+   * as the current one, this is a no-op — Anthropic's stream occasionally
+   * splits a single logical "lane" across multiple blocks (e.g. interleaved
+   * thinking) and we want them visually contiguous within their kind.
+   */
+  function switchStreamBlock(blockType: 'text' | 'thinking' | 'tool_call' | 'tool_result') {
+    if (currentStreamBlockType === blockType && currentStreamText) return;
+    if (blockType !== 'text' && blockType !== 'thinking') {
+      currentStreamBlockType = blockType;
+      return;
+    }
+    currentStreamBuffer = '';
+    currentStreamBlockType = blockType;
+    if (blockType === 'thinking') {
+      currentStreamBuffer = '💭 ';
+      currentStreamText = new TextRenderable(renderer, {
+        id: `stream-thinking-${++messageCounter}`,
+        content: currentStreamBuffer,
+        fg: THINKING_DIM,
+      });
+    } else {
+      currentStreamText = new TextRenderable(renderer, {
+        id: `stream-${++messageCounter}`,
+        content: '',
+        fg: WHITE,
+      });
+    }
+    scrollBox.add(currentStreamText);
   }
 
   function streamToken(text: string) {
@@ -377,6 +413,7 @@ export async function runTui(app: AppContext): Promise<void> {
     streaming = false;
     currentStreamText = null;
     currentStreamBuffer = '';
+    currentStreamBlockType = 'text';
   }
 
   function loadSessionHistory() {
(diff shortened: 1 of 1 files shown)


Anarchid:
## Roast: 🟢 ACCEPTABLE (with 🟡 nits)

Sixty-two lines, one file, additive, type-safe, with a documented "harmless if upstream events never arrive" gating story. You've made it difficult to be insufferable. Have some nits anyway.

### 🟡 MEH

**1. The blockType union is copy-pasted four times** — `src/tui.ts:359, 381, 1206, 1219`

```ts
let currentStreamBlockType: 'text' | 'thinking' | 'tool_call' | 'tool_result' = 'text';
function switchStreamBlock(blockType: 'text' | 'thinking' | 'tool_call' | 'tool_result')
const bt = event.blockType as 'text' | 'thinking' | 'tool_call' | 'tool_result';
const blockType = event.blockType as 'text' | 'thinking' | 'tool_call' | 'tool_result' | undefined;
```

When agent-framework grows a fifth variant (`redacted_thinking` is already on the wire) you'll patch this in four places and miss one. Hoist a `type StreamBlockType = ...` next to the colour constants.

**2. `switchStreamBlock`'s tool_call/tool_result branch updates state and does nothing else — without saying why** — `src/tui.ts:382-385`

```ts
if (blockType !== 'text' && blockType !== 'thinking') {
  currentStreamBlockType = blockType;
  return;
}
```

This is doing real work — it ensures the next `inference:tokens` mismatch-check creates a fresh element when tokens swing back to text/thinking — but a future reader sees "assign and return" and assumes it's dead. The function-level comment talks about same-lane re-entry, not the tool-block bookkeeping. One sentence of comment would save it.

**3. The `💭 ` prefix is hardcoded in two places** — `src/tui.ts:389` and `src/tui.ts:441`

Live stream and `loadSessionHistory` each spell the literal `'💭 '` independently. You already factored out `THINKING_DIM`; `THINKING_PREFIX` is the same one-line constant and avoids a future bug where streaming history looks different from replayed history.

**4. Cast at the trace-event boundary slightly lies** — `src/tui.ts:1206, 1220-1221`

```ts
const bt = event.blockType as 'text' | 'thinking' | 'tool_call' | 'tool_result';
```

The subsequent `if (bt === 'text' || bt === 'thinking')` narrowing makes this operationally safe, so this is taste — but `as string` then the same narrow is more honest about the trust boundary (Anthropic does ship `redacted_thinking`).

**5. No unit test for the new lane-switching state machine**

PR description leans on manual e2e against `data-evac`. Fair — `tui.ts` has zero existing test coverage, so this isn't a regression. But this PR added a small, testable state machine with *two* lane-switch trigger paths (`block_start` and the belt-and-braces `inference:tokens` fallback). A synthetic event stream into a mocked scrollBox would lock in "same-lane re-entry is a no-op" and "tool block sandwich resets the next text element" before they regress six months from now.

### Tooling

```
npx tsc --noEmit          → EXIT=0
bun test                  → 292 pass / 0 fail / 799 expect() calls
```

Dep PRs at time of review: `anima-research/agent-framework#35` OPEN, `antra-tess/membrane#19` OPEN. Consistent with the PR's "correct on its own, gated on upstream" framing.

### Verdict

It is, to my displeasure, fine. Dedupe the union, add a `THINKING_PREFIX` constant, drop a one-line comment on the tool-block branch of `switchStreamBlock`, and ship it after the deps land.

🤖 [/roast](https://claude.com/claude-code) by Claude

Anarchid:
Addressed nits 1–4 in a7f87ae.

1. **Duplicated union** — hoisted `type StreamBlockType` next to the colour constants; the four sites now reference the alias.
2. **`switchStreamBlock`'s tool branch** — added an inline comment explaining that the bookkeeping is load-bearing (it ensures a fresh `TextRenderable` after a tool sandwich rather than reusing the prior element).
3. **`💭 ` duplication** — hoisted `THINKING_PREFIX`; live stream and `loadSessionHistory` now reference it.
4. **Cast honesty** — narrowed the trace boundary to `as string`, kept the subsequent equality narrowing. `redacted_thinking` would now silently fall through rather than masquerade as a known variant.

5. **Unit test for the lane state machine** — point taken, but the function closes over `renderer`, `scrollBox`, and `messageCounter`, so a meaningful test requires either a refactor (extract a pure reducer) or mocking the OpenTUI primitives. There's no precedent for either in the conhost test suite (no `tui.test.ts` exists). Deferred — happy to do the refactor in a follow-up PR if you'd rather have the coverage than the smaller diff here.

Tests still 292/292, tsc clean.

PR #38, opened 2026-06-16
webui: context makeup view (segments + exact tokens)

antra-tess:
Adds a **Context** tab to the webui showing the makeup of the agent's current compiled context.

- **Backend:** `GET /debug/context/makeup` returns the strategy's `RenderStats` (head / middle-raw / summaries L1–L3 / recent tail, each with message + token counts) plus an **exact total** via `count_tokens` (uses a live Claude model id — shared tokenizer, so the count is exact even when the agent's own model is gateway-routed/deprecated). Transparent: `previewActivation` + `count_tokens` only.
- **Frontend:** new SolidJS `ContextPanel` — proportional segment bar + per-segment table + exact total.

Came out of needing to see how much of a long-running agent's context is summarized memory vs raw recent (e.g. ~half L1 summaries at 120k tokens).

### DIFF
--- a/src/modules/web-ui-module.ts
+++ b/src/modules/web-ui-module.ts
@@ -692,6 +692,91 @@ export class WebUiModule implements Module {
     return this.serveStatic(requested);
   }
 
+  /**
+   * Context makeup: the segment breakdown of the agent's current compiled
+   * context — head window, raw middle, summaries by level (L1/L2/L3), and the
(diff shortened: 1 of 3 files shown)


antra-tess:
Closing — based off main, but this feature depends on the /debug/context + previewActivation + RenderStats infra that lives on consolidate/adaptive-default, not main. Will land with that branch.

PR #40, opened 2026-06-23
fleet: unlink stale ipc.sock before spawning a child

Tengro:
## Problem

A fleet child that crashed or was SIGKILLed — e.g. a `docker stop` whose graceful shutdown outran the 30s grace period — leaves its `ipc.sock` behind on the bind-mounted data dir. On the next launch:

1. `waitForSocket()` only checks `existsSync(socketPath)`, so it returns immediately on the **stale** socket.
2. `connectChildSocket()` then races the new child's own stale-socket unlink and fails with **`connect ENOENT`**.
3. The parent SIGKILLs the child — **before it even loads its recipe** — and retries, producing a pre-recipe crash-loop.

Observed in production: the encyclopedist child looped on bare `--- launch ---` markers (no "Loaded recipe") while the container stayed healthy.

## Fix

Unlink the stale `ipc.sock` + `headless.pid` **before spawning** the child, so the only socket `waitForSocket()` can see is the fresh one this child creates. Scoped to the **fresh-launch path only** — the adopt-on-restart path legitimately reconnects to a live child's socket and is untouched.

## Notes

- Pairs with a defense-in-depth sweep in the cook-generated entrypoint (clears stale sockets on container start); this is the in-process fix that also covers mid-run flaps.

### DIFF
--- a/SHERLOCK-COMBINE-PLAN.md
+++ b/SHERLOCK-COMBINE-PLAN.md
@@ -1,0 +1,181 @@
+# Sherlock Combine — Status, Remaining Work, Test Plan
+
+*June 2026. Companion to `HEADLESS-FLEET-PLAN.md` (triumvirate). Spans four
+repos: `zulip_mcp`, `agent-framework`, `forking-knowledge-miner` (host),
+`boter` (wiki sidecar).*
+
+## 1. Concept
+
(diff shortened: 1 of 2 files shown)


Tengro:
Opened against the wrong base repo (connectome-host instead of forking-knowledge-miner) due to a stray upstream remote. Re-opening on the correct repo.

PR #34, opened 2026-06-05
feat(miner): add scribe MCP server for audio/video transcription

dariakroshka:
## Summary

- Adds [scribe-mcp](https://github.com/anima-research/scribe-mcp) as a fifth data source on the knowledge miner, enabling transcription and knowledge extraction from audio/video recordings found on Notion pages
- Extends the miner's system prompt with scribe tool documentation (Section 5) and domain glossary guidance
- Documents Zulip and scribe env vars in `.env.example`

## What scribe provides

Four MCP tools available to the miner as `scribe--*`:

| Tool | Description |
|------|-------------|
| `transcribe` | Transcribe a local audio/video file (Gemini-powered, chunked for long recordings) |
| `probe` | Check file format and estimate token cost before committing |
| `scribe_notion_page` | Scan a Notion page for media, transcribe all attachments, post results as subpages |
| `chat` | Follow-up questions about transcribed content, glossary-aware |

## Setup

Requires three env vars in `.env`:
```ini
GEMINI_API_KEY=...
NOTION_API_KEY=...          # may already exist for syncntn
SCRIBE_GLOSSARY_URL=...     # optional: MediaWiki glossary page URL
```

And the scribe-mcp repo cloned as a sibling: `../scribe-mcp/`

A local glossary file can also be placed at `data/miner/input/glossary.txt`.

## Test plan

- [ ] Miner starts with scribe in `mcpServers` — no crash on recipe load
- [ ] `scribe--probe` returns file info when miner encounters a media file
- [ ] `scribe--scribe_notion_page` transcribes and posts subpages to Notion
- [ ] Miner cites video sources as `[SRC: scribe: <title>]` in reports
- [ ] Miner starts cleanly when `GEMINI_API_KEY` / `SCRIBE_GLOSSARY_URL` are unset (scribe block should be removed from recipe in that case, per existing convention)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #41, opened 2026-07-01
feat(miner): add scribe MCP for audio/video transcription (supersedes #34)

Tengro:
## Summary

Adds [scribe-mcp](https://github.com/dariakroshka/scribe-mcp) as a fifth data source on the knowledge miner — audio/video transcription + knowledge extraction — and documents the `scribe--*` tools in the system prompt (§5).

**Supersedes #34**, which cannot run: its `source.url` `anima-research/scribe-mcp` **does not exist** (404 on clone). Fixed to the real public repo, pinned at `3b5fe5a`.

## Changes

- **`recipes/knowledge-miner.json`** — scribe `mcpServers` block + prompt §5. Source pinned; `install: { runtime: "bun", run: "bun install --frozen-lockfile" }` (scribe ships a `bun.lock` and runs under bun; frozen + pinned = reproducible builds). Prompt describes scribe's **external egress** (uploads media to Google's Gemini API; reads/writes the Notion API) honestly — it does not claim every non-web source is internal.
- **`src/recipe.ts`** — the recipe loader now accepts `'bun'` as an `install.runtime` (alongside `node`/`python3`/`custom`). Required for the recipe above to load. The connectome-cook builder gained matching `bun` support in a parallel change, so `cook build` generates a `FROM oven/bun` builder stage.
- **`.env.example`** — optional `GEMINI_API_KEY` / `NOTION_API_KEY` / `SCRIBE_GLOSSARY_URL`.

Glossary is optional (`SCRIBE_GLOSSARY_URL` unset ⇒ transcribe without it); recipe stays org-neutral.

## Deploy note for cookers

scribe shells out to **ffprobe / ffmpeg / curl** at runtime. A cooked image's runtime stage must include `ffmpeg` and `curl` or every tool but `scribe--chat` fails with ENOENT. (cook does not yet express per-source system packages — tracked as a follow-up.)

## Test plan

- [ ] Recipe loads (loader accepts `runtime: bun`)
- [ ] `cook build` emits a `FROM oven/bun` scribe stage that `bun install --frozen-lockfile` succeeds in
- [ ] With `ffmpeg`+`curl` in the runtime image, `scribe--probe` / `scribe--transcribe` run
- [ ] Miner starts cleanly when `GEMINI_API_KEY` / `SCRIBE_GLOSSARY_URL` are unset


recipes/knowledge-miner.json
```diff
@@ -63,6 +63,28 @@
...
+    "scribe": {
+      "command": "bun",
+      "args": [
+        "../scribe-mcp/src/index.ts"
+      ],
+      "env": {
+        "GEMINI_API_KEY": "${GEMINI_API_KEY}",
+        "NOTION_API_KEY": "${NOTION_API_KEY}",
+        "SCRIBE_GLOSSARY_PATH": "./input/glossary.txt",
+        "SCRIBE_GLOSSARY_URL": "${SCRIBE_GLOSSARY_URL}"
```
Anarchid: **Blocker:** `${SCRIBE_GLOSSARY_URL}` is the *required* interpolation form — `substituteEnvVars()` throws at recipe-load time when it's unset, so the entire miner (all five sources) fails to start, contradicting the "unset ⇒ transcribe without a glossary" behavior documented in the PR body, `.env.example`, and the system prompt itself. Verified by loading this branch's recipe with the var unset.

```suggestion
        "SCRIBE_GLOSSARY_URL": "${SCRIBE_GLOSSARY_URL:-}"
```

The `${VAR:-}` form substitutes an empty string when unset, and scribe's `loadGlossaryFromWiki()` does `if (!url) return undefined`, so an empty string behaves exactly like unset. (Missing `./input/glossary.txt` is already just a stderr warning on scribe's side, so `SCRIBE_GLOSSARY_PATH` is fine as-is.)

recipes/knowledge-miner.json
```diff
@@ -63,6 +63,28 @@
...
       }
+    },
+    "scribe": {
+      "command": "bun",
+      "args": [
+        "../scribe-mcp/src/index.ts"
+      ],
+      "env": {
+        "GEMINI_API_KEY": "${GEMINI_API_KEY}",
+        "NOTION_API_KEY": "${NOTION_API_KEY}",
```
Anarchid: Same issue as `SCRIBE_GLOSSARY_URL` below: Notion is only needed by one of scribe's four tools (`scribe_notion_page`), and scribe creates the Notion client lazily (`notion.ts` throws a clean per-call "NOTION_API_KEY is required" only when that tool is invoked). With the required `${NOTION_API_KEY}` form, an operator without a Notion token loses the entire miner rather than just that one tool.

```suggestion
        "NOTION_API_KEY": "${NOTION_API_KEY:-}",
```

(`${GEMINI_API_KEY}` staying required is right — scribe is useless without it, and fail-at-load matches the syncntn/gitlab convention.)

recipes/knowledge-miner.json
```diff
@@ -63,6 +63,28 @@
...
+        "SCRIBE_GLOSSARY_PATH": "./input/glossary.txt",
+        "SCRIBE_GLOSSARY_URL": "${SCRIBE_GLOSSARY_URL}"
+      },
+      "source": {
+        "url": "https://github.com/dariakroshka/scribe-mcp.git",
+        "install": {
+          "runtime": "node",
```
Anarchid: Non-blocking, but worth recording for the cook path: per `RecipeMcpServerSource` docs, `runtime` tells build tooling which base packages to install in the image — and `node` undersells what scribe needs at runtime. The `command` is `bun` (scribe uses `Bun.file`/`Bun.spawn`), `transcribe` shells out to `ffprobe` unconditionally (`gemini.ts` `getMediaDuration`) and to `ffmpeg` for anything over 10 minutes, and the Notion attachment download uses `curl`. Bun likely survives in a cooked image (the host itself needs it), but the runtime image is minimal — first containerized `transcribe` of a real meeting recording will die at the ffprobe spawn.

Suggest at least noting the ffmpeg/curl dependency here or in the PR description so cook coverage can grow to match; this exact class of recipe-vs-image gap has bitten before.

recipes/knowledge-miner.json
```diff
@@ -1,10 +1,10 @@
...
...any source; added because this is how such systems typically work. Prefer `[WEB]` with a real citation when the claim is publicly checkable; reserve `[GEN]` for background too foundational to bother citing. |\n| `❓` | Knowledge gap — not found anywhere; requires a human expert to fill. |\n\n### Rules for markers\n- **Do not fill gaps with plausible domain logic.** If something is not in the sources, use `❓` and note what needs SME input. Do not invent.\n- **When in doubt, mark it.** An unmarked claim that turns out to be wrong is worse than a `[GEN]` tag that gets verified.\n- **Cite inline, not in footnotes.** A brief tag on the same line is enough: `Max throughput: 250,000 msg/s [SRC: git: config/packetizer]`\n- **Keep internal and external meanings separate.** When a term has both an internal `[SRC]` definition and a public `[WEB]` definition, present them side by side and flag the divergence — don't merge them into one claim.\n- **All written documents are Draft.** Set status to Draft — a human must review and approve before it becomes Active.\n- **Lessons don't need markers.** Lessons have their own confidence scores and evidence arrays. Markers are for prose documents only.",
```
Anarchid: Two prompt nits:

1. The very first sentence still enumerates four sources ("Zulip conversations, Notion pages, GitLab, and the public web") — recordings didn't make it in, even though the "five data sources … four internal" line right after was updated. Wave 1 (Scout) similarly lists per-source scouting actions for everything except scribe.

2. `scribe--transcribe` takes an *absolute* path, but the miner's only view of local files is through workspace mounts (`input/...`) — it has no way to know the host cwd, so the "transcribe a local audio/video file" flow is unreachable as documented (the Notion flow is fine). One sentence in Section 5 mapping the `input/` mount to its on-disk location (`<host cwd>/input/...`) would close the gap.

recipes/knowledge-miner.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "Multi-Source Knowledge Miner",
-  "description": "Extract structured knowledge from Zulip, Notion, and GitLab using forking agents",
+  "description": "Extract structured knowledge from Zulip, Notion, GitLab, and audio/video recordings using forking agents",
   "version": "1.1.0",
```
Anarchid: Nit: a fifth data source + new MCP server + prompt rewrite feels like it warrants a bump to `1.2.0`.

Anarchid (changes requested):
Thanks for redoing this properly — I verified the two claimed fixes over #34 and they hold: `dariakroshka/scribe-mcp` exists, `bun ../scribe-mcp/src/index.ts` matches scribe's own `start` script, tool names in the prompt match the actual `server.tool()` registrations, and the relative paths (`../scribe-mcp`, `./input/glossary.txt`) resolve correctly against the host cwd, consistent with the zulip entry.

One real blocker though, caught by running the recipe through `loadRecipe()`:

```
=== all env vars set ===
LOADED OK — servers: zulip, syncntn, gitlab, ddg, scribe

=== SCRIBE_GLOSSARY_URL unset (the documented-optional var) ===
LOAD FAILED: Recipe references environment variable ${SCRIBE_GLOSSARY_URL} which is not set. ...
```

`substituteEnvVars()` treats bare `${VAR}` as **required** — it throws at load time, taking down the whole miner (all five sources), not just scribe. The optional form is `${VAR:-}` (see `src/recipe.ts:430-433`). Details inline; two one-line fixes make the PR's "glossary is optional" behavior actually true.

Also flagged inline: the `source.install.runtime` declaration vs scribe's actual runtime deps (bun/ffmpeg/curl) for the connectome-cook path, and a couple of prompt/version nits.

Tengro:
## 🔍 Roast review — severity: 🟠 CONCERNING

The diff is small and mostly disciplined — pinned sha, `--frozen-lockfile`, the `bun` runtime guard added to `validateRecipe` in the same PR that needs it. Which makes it all the more disappointing that the one behavioral claim this PR makes in prose is falsified by the loader four hundred lines below it. I checked.

### 🟠 `.env.example` says `SCRIBE_GLOSSARY_URL` is optional. The loader says otherwise.

`.env.example`:
```
# SCRIBE_GLOSSARY_URL=...      # optional: domain-glossary page; unset = transcribe without a glossary
```

`recipes/knowledge-miner.json`:
```json
"SCRIBE_GLOSSARY_URL": "${SCRIBE_GLOSSARY_URL}"
```

`substituteEnvVars` in `src/recipe.ts` (~line 485) is explicit: bare `${FOO}` is **required** and throws if unset; `${FOO:-}` is the documented spelling for "or empty, never throw." Verified against this branch's own loader:

```
$ bun load-test.ts   # all vars set
LOADED OK: Multi-Source Knowledge Miner | servers: zulip,syncntn,gitlab,ddg,scribe

$ bun load-test.ts   # SCRIBE_GLOSSARY_URL unset
error: Recipe "recipes/knowledge-miner.json" references environment variable
${SCRIBE_GLOSSARY_URL} which is not set. Add it to your .env file, supply a
default with ${SCRIBE_GLOSSARY_URL:-...}, or delete the section of the recipe...
```

So "unset = transcribe without a glossary" is actually "unset = the entire miner recipe fails to load and you get no agent at all." The fix is one character pair:

```json
"SCRIBE_GLOSSARY_URL": "${SCRIBE_GLOSSARY_URL:-}"
```

scribe-mcp handles the empty string correctly (`if (process.env.SCRIBE_GLOSSARY_URL)` — falsy on `""`, warns and proceeds glossary-less). The graceful-degradation path already exists on both sides; the recipe just refuses to use it.

### 🟡 Scribe keys are now hard-required for everyone loading this recipe

Same mechanism: bare `${GEMINI_API_KEY}` / `${NOTION_API_KEY}` mean a user who wants to mine Zulip only can no longer load `knowledge-miner.json` without provisioning a Gemini key. That's consistent with how the other sources behave, fine — but then `.env.example` shouldn't present them commented-out like the optional Notion block above them. Say "required if you keep the `scribe` server entry" or don't comment them out.

### 🟡 Host-mode prerequisites: undocumented

The recipe's `command` is `bun ../scribe-mcp/src/index.ts` — so a host-mode operator needs (1) a sibling `scribe-mcp` checkout at the pinned ref, (2) bun, and (3) **ffmpeg/ffprobe/curl on the host**: `gemini.ts:37` spawns `ffprobe` unconditionally on every `transcribe`, and `notion.ts:112` shells out to `curl`. Every other source in this recipe gets setup documentation; scribe gets nothing. The first person to run `scribe--transcribe` on a clean machine gets an ENOENT and a mystery.

### 🟠 Not this PR, but blocking everything downstream of it

`main` @ `58c9db8` (the 0.3.1 release) bumped `package.json` to `@animalabs/agent-framework ^0.6.0` / `context-manager ^0.5.5` **without regenerating `bun.lock`** (still pins 0.5.7 / 0.5.4). Result, on a fresh clone (bun 1.3.10):

```
$ bun install --frozen-lockfile
error: lockfile had changes, but lockfile is frozen
```

Every cook-built image runs exactly that command in its `ch-deps` stage, so no image of connectome-host@main builds today. And after an unfrozen install, `bunx tsc --noEmit` fails too: `src/logging-adapter.ts(62,68): error TS2339: Property 'thinking' does not exist on type 'ProviderRequest'`. Neither is this PR's fault, but this PR's whole purpose is to get cooked into org images — it can't until the lockfile is recommitted.

### Tooling

- Recipe load test (branch loader, above): **pass** with all vars, **throws** with `SCRIBE_GLOSSARY_URL` unset
- `bun install --frozen-lockfile` (fresh clone): **fails** (pre-existing on main)
- `bunx tsc --noEmit` (after unfrozen install): **1 pre-existing error** in `src/logging-adapter.ts`, untouched by this PR
- JSON validity of the recipe: ✅

### Verdict

The scribe integration itself is done the right way — pinned, frozen, guard and consumer shipped together, and the system prompt describes the Gemini/Notion egress honestly instead of pretending it's an internal source. Swap `${SCRIBE_GLOSSARY_URL}` for `${SCRIBE_GLOSSARY_URL:-}` so the `.env.example` stops making promises the loader won't keep, add a line of host-mode setup docs, and this merges with my grudging approval. Fix the lockfile on main separately, and urgently — everything queued behind this PR is stuck on it.

*— automated roast, verified by hand where it matters*

Tengro:
## 🔍 Roast review, round 2 — severity: 🟢 ACCEPTABLE

I came back ready to be disappointed and left mostly unemployed. For the record, re-verified against this branch's own loader:

```
GEMINI set, optionals unset →
  LOADED OK | scribe env: {"GEMINI_API_KEY":"g","NOTION_API_KEY":"",
    "SCRIBE_GLOSSARY_PATH":"./input/glossary.txt","SCRIBE_GLOSSARY_URL":""}
  sysPkgs: ["ffmpeg","curl"]
GEMINI unset →
  throws, exactly as .env.example now says it will
systemPackages: ["ffmpeg", ""] →
  REJECTED — "must be an array of non-empty strings"
```

Every round-1 finding addressed, properly:

- `${NOTION_API_KEY:-}` / `${SCRIBE_GLOSSARY_URL:-}` — the optional claim is now *true*, and scribe's own empty-string handling picks it up from there. ✅
- `.env.example` now states plainly which var is required and why, and names the substitution form for each. This is what honest config documentation looks like. ✅
- `systemPackages` added to the schema with validation *stricter than the consumer's* (rejects empty strings — see my note on cook PR #6, whose vendored copy doesn't). ✅
- The new §5 prompt caveat — "`scribe--transcribe` takes a filesystem path, but your only view of local files is the `input/` mount... prefer `scribe--scribe_notion_page`" — is the kind of failure-mode-anticipating instruction nobody asked for and everybody will silently benefit from. Genuinely good. ✅
- Version bumped, description updated, scout wave mentions media attachments. ✅

### 🟡 The one survivor: host-mode prerequisites still undocumented

`.env.example` covers the env vars now, but a host-mode operator still learns nowhere that they need a sibling `../scribe-mcp` checkout at `3b5fe5a…`, bun, and **ffmpeg/ffprobe/curl on the host** — `gemini.ts` spawns `ffprobe` unconditionally on every transcribe, and `scribe--scribe_notion_page` transcribes too, so the "prefer the Notion path" prompt advice doesn't dodge the binary requirement. One short README/docs paragraph. That's the whole ask. Second time asking.

### Tooling

```
bun install --frozen-lockfile → fails (main's lockfile drift — fixed by PR #42, not this PR's problem)
bunx tsc --noEmit             → 1 pre-existing error (logging-adapter, fixed by PR #42)
bun test                      → 317 pass / 2 fail — the same 2 fleet-e2e tests
                                (fleet-tree-aggregator-e2e, fleet-subscribe-union-e2e)
                                fail identically on origin/main; pre-existing, likely environmental
recipe load matrix            → all green (above)
```

### Verdict

Merge it — **after #42**, which unbreaks the lockfile and tsc that this branch currently inherits from main. Add the host-mode paragraph in a follow-up if you must, but add it. I checked everything twice specifically hoping to find something worse and could not. Well done, I suppose. Don't let it become a habit; I have a reputation.

*— automated roast, round 2, verified by hand where it matters*

Tengro:
## 🔍 Roast review, final round — severity: 🟢 ACCEPTABLE

Verified against this branch's loader, one last time:

```
validateRecipe systemPackages matrix:
  ['ffmpeg','curl','python3-venv','libssl3'] → ACCEPTED
  ['ffmpeg','']                              → rejected
  ['ffmpeg','&& curl evil|sh']               → rejected
  ['FFmpeg']                                 → rejected (Debian names are lowercase)
recipe load (GEMINI set, optionals unset)    → loads; scribe sysPkgs ["ffmpeg","curl"]
regex parity with cook's vendored copy       → byte-identical
bun test                                     → 317/319; the 2 fleet-e2e failures are
                                               identical on origin/main (pre-existing)
```

The round-2 injection concern is closed with the same regex on both sides of the vendoring boundary, in the same batch, with matching cross-references in the comments. That is how you keep two copies of one truth from drifting: you make each one name the other.

**One retraction, for the record.** I asked twice for host-mode setup docs (sibling checkout, bun, ffmpeg). I have now actually checked this repo's documentation norms: the README and docs/ describe *no* source's host-mode setup — not zulip's, not ddg's, not syncntn's. `.env.example` is the established documentation surface here, and scribe's entry in it is now the most honest one in the file. Demanding a bespoke docs section for scribe alone would be inventing a standard mid-review, and the org repo that *does* maintain an install guide (lynx-conhost-recipes) got its section. Withdrawn. It happens roughly once a decade; mark the date.

### Verdict

Three files, and every line of them has now been verified against the loader it feeds, twice. Merge **after #42** (which unfreezes the lockfile this branch still inherits from main). Nothing further from me. Genuinely nothing — I checked.

*— automated roast, final round*

Tengro:
## 🔍 Roast review, round 4 (the systemic one) — severity: 🟢 ACCEPTABLE

The recipe now points at the self-contained scribe (`Tengro/scribe-mcp@c0deddc` — verified to be exactly the head of upstream PR dariakroshka/scribe-mcp#1) and `systemPackages` is gone. Re-verified end to end:

```
loadRecipe (GEMINI set, optionals unset) → LOADED OK
  url: https://github.com/Tengro/scribe-mcp.git | ref: c0deddc | sysPkgs: (none)
scribe block                             → byte-identical with boter !9 and lynx !7 (still)
stale-ref grep (3b5fe5a / dariakroshka)  → clean on this branch
fork @ c0deddc, bun install --frozen-lockfile → binaries land and execute
  (ffmpeg 7.0.2-static + ffprobe, both exit 0 via the resolved paths)
```

Two remarks, neither blocking:

1. **This is the right direction for the schema, too.** `systemPackages` was a correct *capability* and scribe was the wrong *customer* — an MCP that owns its toolchain beats one that exports its dependencies into every consumer's image. The field stays in the loader for sources that genuinely need distro packages; scribe just stopped being one. Both halves of that are correct simultaneously, which the git history now demonstrates rather than argues.
2. **Follow-up when upstream merges:** once dariakroshka/scribe-mcp#1 lands, re-point `source.url`/`ref` to the upstream SHA so the recipe doesn't depend on a personal fork's continued existence. lynx's install.md documents this plan explicitly; a line in this PR's description would carry the same knowledge here.

Merge order is unchanged and unaffected: **#42 first**, then this.

### Verdict

Four rounds ago this PR shipped a recipe whose optional env var could kill the miner. It now ships a pinned, self-contained, loader-verified source block that made two other repos simpler by existing. Nothing to fix. I've re-checked that sentence twice and it survived both readings.

*— automated roast, round 4*

Anarchid (approved):

PR #42, opened 2026-07-02
fix(deps): recommit bun.lock for 0.3.1 bump (fresh clones + cook builds fail) + tsc fix

Tengro:
## Problem

`main` @ `58c9db8` (0.3.1) bumped `package.json` to `@animalabs/agent-framework ^0.6.0` / `context-manager ^0.5.5` **without regenerating `bun.lock`** (still pins 0.5.7 / 0.5.3-4). So on a fresh clone (bun 1.3.10):

```
$ bun install --frozen-lockfile
error: lockfile had changes, but lockfile is frozen
```

Every cook-built image runs exactly that in its `ch-deps` stage, so **no connectome-host image builds today** — which blocks the scribe/boter/lynx work (PR #41 et al.) from ever reaching a container.

## Fix

- **`bun.lock`** regenerated against the declared ranges (resolves af `0.6.0`, cm `0.5.5`, membrane `0.5.68`, chronicle `0.2.1`).
- **`src/logging-adapter.ts`**: after the bump, `bunx tsc --noEmit` fails — membrane 0.6.0's `ProviderRequest` no longer types `thinking`, so the `as unknown as ProviderRequest['thinking']` index fails. The Anthropic adapter still reads a top-level `request.thinking` at runtime (`complete(): if (request.thinking) params.thinking = request.thinking`), so the reasoning injection is unchanged — inject top-level and assert the whole object.

## Verified

- `bun install --frozen-lockfile` → clean
- `bunx tsc --noEmit` → clean

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,9 +5,9 @@
     "": {
       "name": "connectome-host",
       "dependencies": {
-        "@animalabs/agent-framework": "^0.5.7",
+        "@animalabs/agent-framework": "^0.6.0",
         "@animalabs/chronicle": "^0.2.0",
-        "@animalabs/context-manager": "^0.5.4",
+        "@animalabs/context-manager": "^0.5.5",
         "@animalabs/membrane": "^0.5.68",
         "@opentui/core": "^0.1.82",
       },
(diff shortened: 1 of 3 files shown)


Tengro:
## 🔍 Roast review — severity: 🟡 MEH (the fix works; the story it tells about itself doesn't)

This is the keystone PR — nothing in the scribe/boter/lynx train ships until it merges — so I verified it harder than its 20 lines deserve.

### Verified ✅

```
fresh checkout + bun install --frozen-lockfile → clean (124 packages)
bunx tsc --noEmit                              → clean
bun test                                       → 317 pass / 2 fail — same 2 fleet-e2e
                                                 failures on origin/main; pre-existing
```

The lockfile regen is correct: af 0.6.0 / cm 0.5.5, membrane stays 0.5.68, and the now-obsolete nested pins (`agent-framework/context-manager@0.5.3`, its nested `membrane@0.5.55`) are properly dropped. And the runtime claim holds — I read membrane's shipped adapter: `anthropic.js:302` is literally `if (request.thinking) { params.thinking = request.thinking; }`, so the injected top-level field does reach the API params. The fix works.

Now, the three things wrong with how it got there:

### 🟡 The comment blames a membrane version that does not exist

`src/logging-adapter.ts`:
```
// As of membrane 0.6.0 `ProviderRequest` no longer types a `thinking` field...
```

There is no membrane 0.6.0. `npm view @animalabs/membrane versions` ends at **0.5.68** — which is what this very PR's lockfile resolves, and what the PR description *itself* says two paragraphs later ("resolves … membrane `0.5.68`"). The package that went 0.6.0 is **agent-framework**; membrane 0.5.68's `ProviderRequest` (`dist/types/provider.d.ts:110–137`) simply has no `thinking` member. The old `as unknown as ProviderRequest['thinking']` broke because the *dedup* removed whichever stale nested copy used to make that index resolve. A comment that exists to explain a version-drift bug and cites a phantom version is a booby trap for the next person who goes hunting through membrane's changelog for it. Fix the attribution.

### 🟡 You asserted your way around a typed door membrane left open

```ts
return { ...request, thinking: { type: 'adaptive' } } as ProviderRequest;
```

`ProviderRequest.extra` is documented as "Additional provider-specific params", and the adapter applies it right below the thinking check (`anthropic.js:306–308`: destructures out membrane-internal keys, then `Object.assign(params, rest)`). So:

```ts
return { ...request, extra: { ...request.extra, thinking: { type: 'adaptive' } } };
```

— same wire bytes, zero assertions, survives the *next* type reshuffle instead of hiding it, and stops lying to the compiler about what `ProviderRequest` is. The `as ProviderRequest` version compiles precisely because assertions suppress excess-property checking, which is the same mechanism that will let this silently rot later.

### 🟡 An untyped runtime contract with no regression test

The whole point of this fix is that the `thinking` passthrough is now invisible to the type system — the exact category of thing that breaks silently on the next dependency bump, in a repo with a 33-file, 319-test bun suite sitting right there. One test: transform a request through `LoggingAnthropicAdapter`, assert `(result as any).thinking?.type === 'adaptive'`. Had it existed, the original break would have been a red test with a name instead of a tsc error blamed on the wrong package.

### Verdict

Merge it today — it unblocks #41, boter !9, lynx !7, and every cook build on the planet (well, both of them). But fix the comment first; it's two words, and unlike the rest of my complaints, wrong documentation in a *dependency-archaeology* comment actively damages the next investigation. The `extra` refactor and the test can be follow-ups. The fix is right; I just wish it knew *why* it was right.

*— automated roast, verified by hand where it matters*

Tengro:
## 🔍 Roast review, final round — severity: 🟢 ACCEPTABLE (one stale sentence in the description)

All three round-2 findings addressed, and addressed the way I'd have done it, which I note with mild territorial unease:

- **The phantom "membrane 0.6.0"** — the code comment now tells the true story (membrane is 0.5.68, agent-framework went 0.6.0, the old cast compiled only via a stale nested copy that dedup removed). Correct attribution, and it even preserves the archaeology for the next reader. ✅
- **`extra` instead of the assertion** — `{ ...request, extra: { ...request.extra, thinking: { type: 'adaptive' } } }`, zero type assertions. I additionally verified the part that would have made this fix quietly wrong: membrane applies `extra` inside `buildRequest`, which is called from **both** the complete and stream call sites — so the streaming path (the one agents actually use) carries the thinking param too. It does. ✅
- **Regression test** — `test/logging-adapter.test.ts` pins the type-invisible passthrough, with a comment explaining *why* it exists ("so the next break is a red test with a name"). That's a test that will still make sense in a year. ✅

### Tooling (fresh checkout)

```
bun install --frozen-lockfile → clean (124 packages)   ← the whole point of this PR, still true
bunx tsc --noEmit             → clean
bun test                      → 321 pass / 2 fail (323 tests, 34 files — includes the new suite)
                                the 2 failures are the same fleet-e2e tests that fail
                                identically on origin/main; pre-existing, environmental
```

### 🟡 The one leftover: the PR *description* still cites "membrane 0.6.0"

The code comment was fixed; the PR body wasn't — it still opens the Fix section blaming a membrane release that doesn't exist, directly above a Verified section listing membrane `0.5.68`. PR descriptions outlive branches in merge commits and search results. It's an edit box, not a commit; thirty seconds.

### Verdict

Merge it first — it remains the keystone: #41, boter !9, lynx !7, and every cook `ch-deps` stage on earth are queued behind this lockfile. Fix the description on your way past. This went from "misattributed assertion with no test" to "typed, tested, and correctly narrated" in one round, which is the kind of trajectory that makes a reviewer suspicious he's no longer needed. He is, though. Always.

*— automated roast, final round*

PR #46, opened 2026-07-06
feat(recipe): expose per-agent cacheTtl ('5m' | '1h')

LariTesserae:
Two-line companion to anima-research/agent-framework#53: plumbs `recipe.agent.cacheTtl` → `AgentConfig.cacheTtl`, so a recipe can opt a persistent agent into 1-hour prompt-cache TTL.

Rationale + production measurements are in the framework PR (85% of a chatty resident's day-one spend was 5m-TTL cache re-writes; '1h' cuts an equivalent day to ~¼ cost).

Safe to merge in either order: before agent-framework#53 lands, the extra config field is ignored by the Agent constructor (no-op); after, it types and works.

### DIFF
--- a/docs/AGENT-ONBOARDING.md
+++ b/docs/AGENT-ONBOARDING.md
@@ -243,6 +243,7 @@ ln -sfn ../../../connectome-local/context-manager/node_modules/@animalabs/chroni
     "maxTokens": 16384,                     // response cap
     "maxStreamTokens": 180000,              // recompile trigger; keep < model window
     "contextBudgetTokens": 160000,          // SEE §11.1 — MUST fit under (window - maxTokens)
+    "cacheTtl": "1h",                       // prompt-cache TTL; '1h' for residents replying slower than 5m (cache re-write premium dominates spend otherwise), omit/'5m' for rapid loops
     "strategy": {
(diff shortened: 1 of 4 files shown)


Anarchid:
Verified locally: tsc-neutral (same 15 pre-existing MCPL-admin errors as main against published AF 0.6.0 — the spread compiles fine either way, as you noted) and test-neutral (342/2 both sides).

Pushed review follow-ups to the branch before merging:
- `validateRecipe` now rejects a `cacheTtl` that isn't `'5m'`/`'1h'`. Recipes are runtime JSON, so the interface alone doesn't guard the wire — a typo'd `\"1hr\"` would have surfaced as an Anthropic 400 at the agent's first inference, which for a persistent resident is an ugly way to find out. This matches the file's own fail-at-load-time philosophy (see the `thinking` validation).
- Tests for accepted values / unset / rejection.
- `docs/AGENT-ONBOARDING.md` recipe example gains the `cacheTtl` line.

One operational note for anyone reading later: published `@animalabs/agent-framework@0.6.0` predates framework PR #53, so npm-consumed deployments get a no-op knob until the next AF release is tagged. bun-linked dev setups get it immediately.

PR #47, opened 2026-07-07
fix(fragility): lifecycle, fleet/subagent, lessons/retrieval, web, growth caps

Tengro:
Part of the July 2026 fragility-audit remediation (see UNTESTED-ROUTES-TEST-PLAN.md Groups 1–4 + §6.4). ~25 findings across FKM's own code. **422 tests pass (78 new), existing suite fully green; `tsc` shows only the 15 pre-existing `mcpl-admin-module.ts` errors (unpublished agent-framework 0.6.1 API) — zero new.** Version 0.3.1 → 0.3.2 (not published). 5 logical commits by area.

## Fixed (highlights)
**Lifecycle (Group 1):** SIGTERM/SIGINT clean shutdown in all modes (`src/shutdown.ts`); double-start PID-probe guard; 15s shutdown deadline + finally-unlink; `switchSession` rollback + concurrency guard (`src/session-switch.ts`); headless trace rebind after `/session switch`; TUI cleanup deadline + lessons export on all exit paths; **server-side single-use quit-confirm token** (no more any-frame SIGTERM).

**Fleet/subagent (Group 2):** flap-cap restart history survives respawn; fork sibling-`tool_use` stripping (no orphaned tool_use → 400); typed-first error classification (`\b`-anchored codes, `rate[ _-]?limit` — the "~195023 tokens"/"failed to generate" false positives are tested); slot double-release fix; `returnedResults` keyed by unique name + cleared on all terminal paths; dead-parent delivery gating; adopted-child crash poll; handleLaunch synchronous name reservation; aggregator latch timeout + subtree reset; work-timeout no longer retried as transient.

**Lessons/retrieval (Group 3):** read-merge-write + atomic tmp/rename + `.corrupt-<ts>` backup (no cross-process clobber); full-UUID IDs; clamped confidence; `contextTimeoutMs=15s` (consumes the agent-framework contract, inert on 0.6.0), empty-result cache served, generation-counter cancellation; fail-**closed** on parse failure; debounced async writes.

**Web (Group 4):** protocol-version freeze/banner; monotonic trace seq + welcome-window buffer; backpressure check + drain; prune timer; capped lessons frame.

**Growth (§6.4) + activity:** llm-calls log 10MB rotation; callIdIndex eviction on terminal tool events; TUI map caps; typing indicator cleared on failed/exhausted/aborted inference.

## Partial / deferred (documented)
- 2.6 full index rekey by collision-proof name — narrow ownership-check landed; full rekey deferred (ripples through peek/TUI/cancel APIs).
- 3.1 true cross-**process** lockfile — read-merge-write greatly narrows it; documented residual millisecond window remains.
- 2.12 remaining magic-number config knobs; 4.4 SPA "showing 500 of N" affordance.
- subscription-gc agent filter and channel-mode reconciliation — not attempted (kept out of the way of the [H] work).

Note: `src/modules/mcpl-admin-module.ts` references the unpublished agent-framework 0.6.1 API — its 15 tsc errors predate this branch and clear once agent-framework is published.

---

## Review follow-up (roast round 2)

Addressed in `578b4a8` + `b334204`:
- **[must-fix] TUI hung-stop now force-exits** — the cleanup path discarded the `stopWithDeadline` outcome, so a hung `framework.stop()` unblocked the exit await but left MCPL/fleet/web-ui handles pinning the event loop (`docker stop` rode out to SIGKILL). New `finalizeShutdown` force-exits(1) on a non-clean outcome, matching headless/piped; a clean stop keeps the natural terminal-restoring return. Regression tests added.
- Reset `corruptBackupDone` on clean parse (2nd corruption is backed up); drop the `returnedResults` display-name fallback (callerAgentName is always set on the dispatch path); dead-letter dropped async results to `dropped-results.jsonl`; atomic `{flag:wx}` pid acquire (TOCTOU); stable `llm-calls.jsonl` (directory-bounded across restarts); one shared `SHUTDOWN_DEADLINE_MS`; `restartHistory` cleanup on kill/restart/stop; de-NUL `web-ui-protocol.test.ts`. +11 tests (433 pass).\n\n**Cross-PR coordination (agent-framework #54):** this PR makes the graceful `framework.stop()` run on every TUI/piped shutdown. That is safe *because* #54 was updated to remove MCPL checkpoint destruction from the graceful `stop()` path (permanent removal is now owned by `disconnectMcplServer`, not the close handler). A code comment at the TUI shutdown wiring records this dependency.


Tengro:
# 🔥 Roast Review (automated, Claude)

**Overall severity: 🟡 MEH** — and I want to be clear that this is, by the standards of what I'm usually handed, praise. 422 tests green, 78 new ones that actually exercise the failure modes they claim to (I checked — the flap-cap test really does simulate the delete/recreate cycle, the TOCTOU test really does race two launches), zero new type errors. The findings below are real, but most of them are second-order holes in fixes that are otherwise correct. The one that stings is that the flagship shutdown-deadline fix quietly doesn't deliver in exactly one of the three modes it was applied to.

---

## Tooling results

`npx tsc --noEmit` — exactly the 15 pre-existing `src/modules/mcpl-admin-module.ts` errors (unpublished agent-framework 0.6.1 API), zero new. Matches the PR claim.

`bun test` at PR head (`66c8f70`):

```
 422 pass
 0 fail
 1144 expect() calls
Ran 422 tests across 46 files. [36.90s]
```

Fully green, including the fleet e2e suites.

---

## 🟠 CONCERNING

### 1. The TUI shutdown deadline resolves a promise nobody acts on — a hung `stop()` still parks the process

`src/tui.ts:2105`:

```ts
void stopWithDeadline(() => app.framework.stop(), 15_000).finally(() => {
  resolveExit?.();
});
```

Compare your own siblings. Headless (`src/headless.ts`): `process.exit(outcome === 'stopped' ? 0 : 1)`. Piped (`src/index.ts:626`): `process.exit(outcome === 'stopped' ? 0 : 1)`. TUI: the outcome is **discarded**, `resolveExit` resolves, `runTui` returns, `main()` returns (`src/index.ts:880-882`)… and then nothing. There is no `process.exit` on this path. If `stop()` timed out, the very things that made it hang — MCPL stdio children with live pipes, fleet child sockets, the web-ui `Bun.serve` listener — are still holding the event loop open. The deadline fires at 15s, resolves a promise, and the process parks anyway.

So audit 1.6 is fixed for the *rejecting* `stop()` (the `.finally` handles that, genuinely) but not for the *hanging* `stop()` — which is the case the comment on line 2104 explicitly claims to bound ("the deadline bounds a hung stop()"). It bounds the promise. It does not bound the process. `docker stop` sends exactly one SIGTERM; your second-signal force-exit never gets its second signal, and the container rides out its grace period to the SIGKILL, same as before the fix.

**Fix**: match your other two modes:

```ts
void stopWithDeadline(() => app.framework.stop(), 15_000).then((outcome) => {
  resolveExit?.();
  if (outcome !== 'stopped') process.exit(1);
});
```

And there's a regression-test asymmetry that let this slip: `test/headless-lifecycle.test.ts` drills SIGTERM for headless and piped modes, but nothing drills the TUI hung-stop path.

### 2. Cross-PR interaction: this PR turns agent-framework #54's checkpoint destruction from latent into every-shutdown

Not your bug, but your amplifier. The agent-framework PR #54 review found that graceful `stop()` destroys MCPL checkpoints. Before this PR, SIGTERM in TUI/piped modes hit the runtime default handler — instant death, `framework.stop()` never ran, and the checkpoints were *accidentally preserved*. Now (`src/index.ts:335-348`, `src/tui.ts:2115-2119`) every `docker stop`, every Ctrl+C, dutifully runs the graceful path — and with it, the checkpoint destruction, on every single shutdown. Until #54's `stop()` behavior is fixed, this PR makes shutdowns *more* destructive for MCPL state, not less. Worth a coordination note in the PR body at minimum; ideally land the #54 fix first.

---

## 🟡 MEH

### 3. `corruptBackupDone` is one-shot per process lifetime — corruption #2 gets silently clobbered, which is the exact hole this fix claims to close

`src/modules/lessons-module.ts` (`readGlobal`):

```ts
} catch (err) {
  if (this.corruptBackupDone) return null; // already backed up + logged
```

The flag means "some corruption was backed up once, ever", not "this corruption was backed up". Sequence: file corrupts at t0 → backed up, loudly logged, then overwritten with valid state by the next save; flag stays `true` for the life of the process. File corrupts *again* at t5 (older-version writer, manual edit, disk hiccup) → `readGlobal` returns `null` **silently**, and `saveToGlobal` proceeds to `renameSync` right over it. No backup, no log. The comment block above this function promises "never silently overwritten"; the flag breaks that promise on the second incident. **Fix** is one line: reset `this.corruptBackupDone = false` on any successful parse — then the flag correctly means "this ongoing incident is already backed up" and nothing more.

### 4. The `returnedResults` display-name fallback keeps audit 2.5 alive for exactly the callers it was meant to protect

`src/modules/subagent-module.ts` (runSpawn/runFork):

```ts
const returned = this.returnedResults.get(agentName) ?? this.returnedResults.get(input.name);
...
} finally {
  ...
  this.returnedResults.delete(agentName);
  this.returnedResults.delete(input.name);
```

When `callerAgentName` is present (the normal framework dispatch path), the unique-key fix is correct and tested. But the fallback path re-creates the original bug wholesale: with two same-named runs whose returns landed under the display name, run 1's consumption can eat run 2's stash, and run 1's `finally` unconditionally deletes the `input.name` key — destroying run 2's pending result either way. If `callerAgentName` is genuinely always set in production, say so and drop the fallback deletes; if it isn't, the 2.5 fix has a documented-in-passing hole. Half-keyed maps are how we got here.

### 5. The pid-probe double-start guard has a TOCTOU, and the comment overclaims

`src/headless.ts:115` probes, `src/headless.ts:139` writes the pid file — with a plain `writeFileSync`, ~20 lines apart, no `wx` flag. Two instances started simultaneously both read no-live-pid, both unlink the socket, both proceed — the exact double-writer scenario audit 1.2 set out to prevent, just with a narrower window. Meanwhile the updated comment at line 99 asserts "the PID probe above cleared the 'still alive' case, so anything left here is genuinely stale and safe to remove", which is only true single-threadedly. `writeFileSync(pidPath, ..., { flag: 'wx' })` with a probe-on-EEXIST retry closes it properly. Also worth a code comment: a recycled PID (stale file, number now owned by an unrelated process — EPERM counts as alive per `readAlivePid`) permanently refuses startup until the operator deletes the file. The error message is good enough to live with that; the behavior just shouldn't be a surprise found in production.

### 6. The llm-calls rotation caps the file, not the directory — because the filename is per-boot

`src/logging-adapter.ts` caps a log at 10MB + one `.1` — 20MB, fine. But `src/index.ts:771-774`:

```ts
const llmLogPath = join(
  config.dataDir,
  `llm-calls.${new Date().toISOString().replace(/[:.]/g, '-')}.jsonl`,
);
```

Every boot mints a fresh timestamped file, and nothing ever deletes old ones. On a flapping host — the failure mode half this PR is about — audit 6.4's "tens of GB" becomes "20MB × N restarts", which is smaller but still unbounded, in the growth-caps commit. Either use a stable filename (the rotation you just built handles it) or sweep `llm-calls.*.jsonl` older than the last K boots at startup.

### 7. Dead-parent async results are dropped, not dead-lettered

`src/modules/subagent-module.ts` (`deliverAsyncResult`): the gating itself is right — polluting the root context was worse. But the `console.error` records only *that* a result was dropped, not the result. That summary might be ten minutes of Opus at 165K context. A dead-letter append (`dataDir/dropped-results.jsonl`, through your shiny new `RotatingJsonlLog`, even) costs five lines and makes the work recoverable instead of gone.

### 8. The 15s shutdown deadline is now three independent copies

`SHUTDOWN_DEADLINE_MS = 15_000` in `src/headless.ts:73`; inline `15_000` in `src/index.ts:340`; inline `15_000` in `src/tui.ts:2105`. This PR's own commit trail cites audit 2.12 — "two independent copies of the same literal" — as a defect it fixed for `ZOMBIE_THRESHOLD_MS`. Then it committed three copies of a new literal. Export one constant from `src/shutdown.ts`, which exists precisely to be this file.

### 9. `test/web-ui-protocol.test.ts` is binary to git — this PR's new tests in it are unreviewable

The file contains a raw NUL byte (line 466, a `'a\0b'` literal in a validation test), so git and GitHub render the change as `Binary files differ` — I had to diff it manually to confirm the new `isProtocolCompatible` tests are fine (they are). The NUL predates this branch, but this PR chose to add tests to a file no reviewer can see in the PR diff. Escape it as `'ab'` and the whole file returns to reviewability.

### 10. Small growth/cleanup residue, in the growth-caps PR

- `restartHistory` (`src/modules/fleet-module.ts:246`) is only ever `get`/`set` — never deleted on child kill/removal or `stop()`. Bytes-trivial, principle-relevant. (Credit where due: `adoptedPollTimers` is cleaned on every path I checked.)
- `FleetTreeAggregator.resetChildState` sets `hasInitialSnapshot = false`, which disables the `eventTs < lastSnapshotTs` gate — any straggler frames flushed after `lifecycle:exiting` re-create ghost nodes in the just-cleared tree until the next snapshot wipes them. Narrow window, self-healing, but the reset makes the tree *more* accepting of stale events, not less.

---

## Cross-checks against the sister PRs (requested)

- **fork PR #3 stale `ipc.sock` unlink**: properly superseded. The child now pid-probes before unlinking its own stale socket (`src/headless.ts:115-131`), and the parent's `cleanupStaleChildFiles` covers the adopt path. Good.
- **context-manager PR #32** (`enforceToolPairing` only wired into `selectHierarchical` while FKM defaults `adaptiveResolution: true` at `src/index.ts:432-433`): this PR doesn't touch strategy config — the gap is neither addressed nor worsened. Still open on the context-manager side.
- **agent-framework PR #54**: see finding 2; the interaction is the concern, not overlap.

---

## Verdict

This is what a fragility PR should look like: every fix has a regression test that fails on the pre-fix behavior, the error-classification tests pin the actual false positives from the postmortem ("~195023 tokens", "failed to generate"), and the growth caps mostly degrade gracefully (`capLessons` shipping `truncated`/`total` instead of silently lying is exactly right). Fix the TUI exit path — it's a five-line change and it's the difference between the headline Group 1 fix being true or being true in two of three modes — reset the corruption flag, and coordinate the #54 stop() interaction before merge. The rest is polish. I've seen far worse merged with applause.

Tengro:
## 🔥 Roast Re-Review (automated, Claude)

**Re-review of PR #47 at head `b334204` (`fix/fragility-audit-jul2026`).** Prior review was at `66c8f70`; this pass covers the two fix commits `578b4a8` ("address PR #47 review") and `b334204` ("unique-key result stash + dead-letter"), with whole-PR context retained.

### Overall severity: 🟡 MEH — and I want to be clear that this is the *good* kind of MEH

Eight of ten prior findings are properly fixed, most with regression tests that test the actual failure mode rather than the happy path adjacent to it. What remains is one incorrect detail inside an otherwise-correct fix (the pid-file EEXIST handler un-fixes its own race), one gap in the new dead-letter path, and assorted nits.

---

### Prior findings — verdicts

| # | Prior finding | Verdict |
|---|---|---|
| 1 | 🟠 TUI discards `stopWithDeadline` outcome → hung `framework.stop()` parks the process | ✅ **FIXED** — new `finalizeShutdown()` in `src/shutdown.ts` force-exits(1) on `timed-out`/`stop-failed` after unblocking the await; `src/tui.ts:2118` uses it. Four unit tests in `test/shutdown.test.ts`, **including a wedged `stop()`** (`new Promise(() => {})`) and a rejecting one, with an injectable `exit` to observe the decision. This is exactly the regression test I asked for. |
| 2 | 🟠 Cross-PR: graceful `stop()` on every signal amplified agent-framework #54's checkpoint destruction | ✅ **RESOLVED (externally) + documented** — #54 fixed the close-handler checkpoint removal (verified at its head `3c35f7db`), and this PR now carries an explicit cross-PR NOTE at the tui.ts call site naming the dependency ("Depends on agent-framework #54 having removed MCPL checkpoint destruction from the graceful stop() path"). Not version-pinned, but the tripwire is written down where the next person will trip. |
| 3 | 🟡 `corruptBackupDone` one-shot per process — second corruption silently clobbered | ✅ **FIXED** — flag resets on successful parse (`lessons-module.ts:493`), exactly as suggested ("this ongoing incident is already backed up", not "corruption happened once"). Regression test drives corrupt → heal → corrupt-again and asserts **two** distinct `.corrupt-*` backups with the right bytes and two loud logs. No backup churn: a persistently-corrupt file still backs up once per incident (flag only resets on a *clean* read). |
| 4 | 🟡 `returnedResults` display-name fallback re-created cross-contamination | ✅ **FIXED** — `?? this.returnedResults.get(input.name)` fallback and the `delete(input.name)` are gone in both `runSpawn` and `runFork`; stash is keyed solely by the unique framework agent name. I verified the load-bearing claim: `framework.ts:3535` (`dispatchToolCall`) does enrich every call with `callerAgentName: agentName`. |
| 5 | 🟡 pid-probe→pid-write TOCTOU; recycled-PID permanent block | ⚠️ **PARTIAL** — `{ flag: 'wx' }` atomic acquire added, stale-dead-pid reclaim has a real subprocess test (`headless-lifecycle.test.ts`), recycled-PID caveat documented as accepted. **But the EEXIST handler re-opens the very race it claims to close** — see new finding 1 below. |
| 6 | 🟡 Per-boot timestamped `llm-calls.<ts>.jsonl` → unbounded dataDir on flapping hosts | ✅ **FIXED** — stable `llm-calls.jsonl` (`index.ts`), and the test simulates 8 boots and asserts the directory holds at most `path` + `path.1` bounded by ~2× the cap. |
| 7 | 🟡 Dead-parent async results dropped with only `console.error` | ✅ **FIXED (with one gap)** — `dropped-results.jsonl` dead-letter via `RotatingJsonlLog` (lazily created, 10MB + `.1` cap so the DLQ can't itself become the growth problem, never throws), wired for both results and errors, `dataDir` plumbed from `index.ts`. Test asserts the drop is still gated from the live context but recoverable on disk. Gap: the `!this.ctx` early return skips it — new finding 2. |
| 8 | 🟡 15s shutdown deadline duplicated 3× | ✅ **FIXED** — `SHUTDOWN_DEADLINE_MS` exported once from `shutdown.ts`, all three call sites import it. |
| 9 | 🟡 Raw NUL byte in `test/web-ui-protocol.test.ts` → binary diff | ✅ **FIXED at head** — the NUL is now the escape sequence `'a\x00b'` (line 153); the head blob contains zero NUL bytes (byte-verified). The PR diff for this file *still* renders "Binary files differ" because the **base-side** blob on main (`862a418`) contains the raw NUL — inherited, not this branch's doing, and it self-resolves once this merges. |
| 10 | 🟡 `restartHistory` never evicted; aggregator reset drops stale-event gate | ⚠️ **PARTIAL (eviction fixed; gate accepted with rationale)** — history now cleared on `handleKill`, `handleRestart`, and wholesale on `stop()`, each with tests. The aggregator ghost-node window is explicitly WONTFIX'd with a documented reason I'll actually accept: `lastSnapshotTs` is a child-clock value, and re-arming the gate with parent `Date.now()` would mix clock domains and trade a transient, self-healing ghost for silent event loss under skew. That's a real trade-off analysis written where it belongs, not a shrug. |

---

### New findings (in the fix commits)

#### 🟡 1. `src/headless.ts:130-134` — the EEXIST handler unlinks *before* re-probing, so the concurrent-start race survives its own fix

```ts
if ((err as NodeJS.ErrnoException).code === 'EEXIST' && attempt < 3) {
  // File exists but probed as dead (or appeared in the create race).
  // Reclaim it and retry; a live owner is caught by readAlivePid next loop.
  try { unlinkSync(pidPath); } catch { /* another racer may have won it */ }
  continue;
}
```

The comment is wrong on its own terms: "a live owner is caught by readAlivePid next loop" — no, it isn't, because you just **unlinked the live owner's pid file**, so the next loop's `readAlivePid` reads nothing, returns `null`, and the `wx` write happily succeeds.

Walk the exact scenario audit 1.2 exists for — two instances started simultaneously on one data dir:

1. A and B both probe: no pid file, both get `null`.
2. A's `wx` write succeeds. A owns the lock and proceeds.
3. B's `wx` write → `EEXIST`. B **unlinks A's live pid file** and continues.
4. B re-probes: file is gone → `null`. B's `wx` succeeds. B proceeds.

Result: **both instances run** — two writers on the same append-only Chronicle store, the precise undefined behavior this guard was built to prevent — and as a bonus, A's pid record has been destroyed and replaced by B's, so every subsequent probe sees only B. For the simultaneous-start case this is functionally the same double-start as the old probe-then-write, wearing an "ATOMICALLY" comment.

The stale-dead-pid case (the common one) *is* correctly fixed and tested. But the fix for the race case has the operations in the wrong order.

**Fix:** on `EEXIST`, re-probe *first*; only unlink if the owner is dead:

```ts
if ((err as NodeJS.ErrnoException).code === 'EEXIST' && attempt < 3) {
  const owner = readAlivePid(pidPath);
  if (owner !== null) {
    throw new Error(`refusing to start: another headless instance (pid ${owner}) ...`);
  }
  try { unlinkSync(pidPath); } catch { /* racer */ }
  continue;
}
```

With that order, step 3 becomes: B re-probes, reads A's live pid, refuses. (There remains a microscopic window where B reads the file between A's `open(O_CREAT|O_EXCL)` and A's content write and parses empty → treats as dead; if you care, retry-with-delay before unlink. I won't die on that hill. I will die on the unlink-before-probe one.)

#### 🟡 2. `src/modules/subagent-module.ts:1657-1658` — the dead-letter is skipped on the path where results are most likely to drop

```ts
private deliverAsyncResult(name: string, result: SubagentResult, parentAgentName: string): void {
  this.asyncHandles.delete(name);
  if (!this.ctx) return;          // ← silent drop: no dead letter, no console line

  if (!this.isParentAlive(parentAgentName)) {
    ...
    this.deadLetter({ kind: 'result', ... });
```

`ctx` is nulled when the module stops. So an async subagent whose completion lands *after* shutdown began — which is exactly when in-flight results become undeliverable, and exactly the "ten minutes of Opus at 165K context" the dead-letter comment worries about — hits the `!this.ctx` return and vanishes with **neither** a dead letter **nor** the console line the dead-parent path at least gets. The PR built a recovery mechanism and then routed the most probable loss path around it. Same shape in `deliverAsyncError`.

**Fix:** `deadLetter()` needs only `config.dataDir`, not `ctx` — dead-letter (with e.g. `kind: 'result', reason: 'module-stopped'`) before the `!this.ctx` return, or inside that branch.

#### Nits (no action demanded, noting for the record)

- **`finalizeShutdown` is used exactly once.** `forceExitOnClean: true` exists to express headless/piped semantics, but `src/headless.ts:559` and `src/index.ts:627` still run their own inline `stopWithDeadline` + `process.exit`. So the parameter's only production-adjacent exerciser is its own unit test. Either migrate the other two call sites onto the abstraction you built for them, or accept that you wrote a nice function with a decorative option.
- **`subagent-module.ts:881-884`** — when `callerAgentName` is somehow absent, the `return` handler discards the result and still replies `{ success: true, data: 'Result received.' }`. The comments argue convincingly that the key is always present in production; fine — then the impossible branch deserves a `console.error`, not a polite lie to the subagent.
- **`lessons-module.ts:497`** — `.corrupt-<Date.now()>` backups are unbounded across repeated incidents, in a PR whose banner theme is growth caps. Each one requires a genuine fresh corruption event, so I'll allow it — but a flaky disk on a long-lived daemon will paper the dataDir.

---

### Tooling results

**`npx tsc --noEmit`** — the same 15 pre-existing errors in `src/modules/mcpl-admin-module.ts` (missing `McplServerConfig` export / `listMcplServers` / `restartMcplServer` / `connectMcplServer` / `disconnectMcplServer` on `AgentFramework`, plus implicit-`any` fallout), identical to the prior head. **No new type errors.**

**`bun test`** —

```
 433 pass
 0 fail
 1176 expect() calls
Ran 433 tests across 46 files. [37.38s]
```

Prior head was 422/422; the 11 new tests are the regression coverage described above (hung/rejecting/clean `finalizeShutdown`, stale-pid reclaim, second-corruption backup, multi-boot log rotation bound, restartHistory eviction ×2, dead-letter).

---

### Verdict

Last time I had one 🟠 that could park a container until SIGKILL and a spread of 🟡s. This round: the 🟠 is genuinely fixed — with the injectable-exit seam test I asked for, not a vibes test — the cross-PR dependency is written down at the call site, and most of the 🟡s got fixes *with tests that reproduce the original failure*. The two things I'm sending back are small but real: the pid-file EEXIST handler performs its unlink and its liveness check in the wrong order, which quietly re-admits the double-writer scenario the whole guard exists for; and the dead-letter queue has a bypass lane on the shutdown path it was arguably built for. Fix the ordering, move one call above one early-return, and I'll have nothing left to be weary about — which, for this codebase, I'd almost find unsettling.

PR #48, opened 2026-07-10
Recipe passthrough for productionBudgetTokens

LariTesserae:
Expose Context Manager's standing `productionBudgetTokens` target through current Connectome Host recipe composition.

This is a **context-token production target**, not a provider-spend ceiling: it asks autobiographical maintenance to keep the summary forest deep enough for a later live-budget descent without a fold storm.

## Changes

- add `productionBudgetTokens?: number` to `RecipeStrategy`;
- pass an explicitly configured value through `buildFrameworkStrategy`;
- preserve omission as omission;
- focused test pins both configured and omitted behavior.

## Verification

Current-main rebase at head `a2d1160`:

- 604 passed
- 0 failed
- 1,942 assertions
- 68 files

Local Bun 1.3.14 proposed a lock rewrite during independent setup, but the PR's authoritative Ubuntu and macOS CI both completed their frozen install and full test path successfully; no lockfile change was needed for this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- **Standing autobiographical production target.** Recipes may set `agent.strategy.productionBudgetTokens` to keep the summary forest deep enough for a later live context-budget descent without a fold storm. This is a context-token target passed through to Context Manager, not a provider-spend ceiling; omission preserves Context Manager defaults.
+
 - **`BEDROCK_BASE_URL` env hook** for the bedrock provider — mirrors
   `ANTHROPIC_BASE_URL`, routing bedrock-runtime calls through an inference
   gateway (gate.animalabs.ai/bedrock/<credSet>). The gate reads the agent
--- a/src/framework-strategy.ts
+++ b/src/framework-strategy.ts
@@ -17,6 +17,7 @@ const PASSTHROUGH_KEYS: ReadonlyArray<keyof RecipeStrategy> = [
   'targetChunkTokens',
   'mergeThreshold',
   'summaryTargetTokens',
+  'productionBudgetTokens',
   'l1BudgetTokens',
   'l2BudgetTokens',
   'l3BudgetTokens',
--- a/src/recipe.ts
+++ b/src/recipe.ts
@@ -55,6 +55,10 @@ export interface RecipeStrategy {
   targetChunkTokens?: number;
   mergeThreshold?: number;
   summaryTargetTokens?: number;
+  /** Standing production target: keep the summary forest deep enough to fit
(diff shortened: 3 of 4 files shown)


Anarchid:
Status: the CM side landed on context-manager main (in-house redo of anima-research/context-manager#34, commit 4751a1d — see the close-out comment there for the deltas; you're credited as co-author). This passthrough is now waiting on a CM npm release; after that it needs a rebase (the passthrough list and `RecipeStrategy` both moved since Jul 10) and it can go in. No action needed on your end unless you want to do the rebase yourself.

antra-tess:
## Sol review — useful, but refresh before merge

**REVISE / rebase, then likely accept.** The feature remains useful: current Context Manager still exposes and consumes `productionBudgetTokens` as the standing shadow-pick target, and current Host main still does not pass it through.

The five-line change itself is directionally correct, but this head (`987d964`) is based on the old `connectome-host@0.3.1` tree while current main is `@animalabs/connectome-host@0.7.4`; the recipe and composition surfaces have moved substantially, and this PR has no current CI/check receipt. I don't want a July baseline merged into August composition on the strength of an old “same one failure as main” note.

Please:

1. rebase onto current `main`;
2. retain the typed `RecipeStrategy.productionBudgetTokens?: number` field and current strategy passthrough at the modern construction site;
3. add one focused recipe/composition test proving a configured value reaches the Context Manager strategy options unchanged (and omission remains omission);
4. run the current Host suite.

Nonblocking documentation note: this name is a **context-token production target**, not a provider-spend ceiling; keep that distinction explicit because “production budget” otherwise reads like a cost limit.

Once rebased with the focused pass-through proof and green current suite, I expect this to be mergeable.

antra-tess:
## Sol re-review — ACCEPT

I took over the orphaned branch, rebased it onto current Host, moved the passthrough to the current `buildFrameworkStrategy` composition surface, and added a focused proof that an explicit value reaches Context Manager unchanged while omission stays omitted.

Exact head `a2d1160`: **604 passed, 0 failed, 1,942 assertions across 68 files.**

**ACCEPT / safe to merge.** The field remains explicitly a context-token production target, not a spend ceiling.

antra-tess:
Final head `6a7f485`: Ubuntu CI, macOS CI, and changelog gate all green. Prior 604/604 independent suite remains the code-path receipt; the only subsequent change is the required Unreleased entry. ACCEPT remains.

antra-tess:
Correction to my earlier local setup note: the final PR’s Ubuntu and macOS CI both passed their authoritative frozen-install path. My Bun 1.3.14 checkout merely wanted a local lock rewrite; that is not evidence of a repository lock defect. No Host lock issue is owed.

PR #60, opened 2026-07-20
Operator-facing observability: webui ops/health/branches panels + TUI deep peek and alerts

Tengro:
## What

The July observability work (quarantine klaxon, refusal dragnet, per-agent runtime settings, `/healthz` exposure) made failure states visible to *machines* — failures.log, the ops webhook, connectome-doctor. This PR makes the same states visible to the *human operator* in both UIs, and closes the two long-standing "peek isn't honest" / "branches are invisible" gaps.

### Web UI

- **Ops-alert strip**: `ops:alert` traces render as persistent banner rows (compression quarantine, refusal streaks, inference-exhausted). `<kind>-clear` stands the matching alert down; repeat firings fold into one row with a ×N counter. Dismissal is soft — re-fires reappear.
- **Health sidebar tab**: polls `/healthz` every 15s — per-agent status, consecutive-failure streaks, last error, refusal stats, runtime settings (budget/tail/pace/think-text policy/convergence), compression quarantine detail. Durable-state alerts are **reconciled on connect**, so a page opened mid-incident alarms immediately instead of waiting for the next 15-min klaxon re-fire. Observer grants without the `health` scope get a scope hint.
- **Branch lineage panel**: the header branch chip (now always visible, incl. `main`) opens a Chronicle lineage tree — fork ancestry from `parentId`/`branchPoint`, fork-point markers, per-row checkout. Checkout routes through the existing `/checkout` command path so undo/redo bookkeeping and `_config` re-materialization stay on the one audited path. `branch-changed` frames now update the client (header chip + panel refresh) instead of being handled as no-ops.
- **Protocol**: additive v1 frames `request-branches` → `branches-list` (same precedent as the observer frames). Observer read access rides the `messages` scope; observers remain read-only (no checkout affordance).
- Cleanups: `/curve` link surfaced in the Context panel header, dead `PlaceholderPanel` removed, stale `WEBUI-PLAN.md` / knowledge-miner doc pointers fixed, README rewritten to lead with the web UI.

### TUI

- **Honest per-agent peek into fleet children**: `p` on an agent row inside a fleet child opens that agent's stream — the child's IPC events filtered by `agentName` — rather than dumping the whole process stream. Works for the child's root agent and its subagents (sub-subagents from the parent's seat). Header shows phase/elapsed/tool-calls/ctx-tokens/task/parent from the tree reducer. Filtered and whole-process logs are keyed separately (`child#agent` vs `child`) so they never mix. Honest by construction: the fleet IPC forwards child framework traces verbatim with `agentName` intact (default subscription `'*'`, so `inference:tokens` flows).
- **Ops alerts in the terminal**: `ops:alert` from the local framework *and every fleet child* prints a red chat line (`⚠ [clerk/clerk] compression-quarantine: …`, ×N fold-up) and pins a `⚠ N alerts` segment in the status bar (bar turns red) until the `-clear` stands it down (cyan announcement). Alerts are cleared on session switch — the old session's all-clear can never arrive, and the new session's klaxons re-fire within their alarm interval if the condition holds.
- Session cost estimate (`$0.85`) on the status token line, from the UsageTracker totals.

## Testing

- `tsc --noEmit` clean (host + web), `vite build` clean, `bun test`: **407 pass / 0 fail** (50 files)
- New coverage: `request-branches` in the protocol type-guard suite; e2e assertion that a `health,ops`-scoped observer is refused `request-branches` (rides `messages`)
- TUI smoke-booted under a pseudo-TTY with isolated `DATA_DIR`: 12s alive, alternate-screen rendering, empty `tui-error.log`
- Review pass fixed three defects before submission: stale ops alerts surviving session switches (TUI), `at`-timestamp churn in the health reconciler causing perpetual "5s ago" + banner reshuffling (SPA), and unguarded recursion over a hypothetically-cyclic branch lineage (SPA)

## Not in scope

Call-ledger and branch views in the TUI (web-only for now), cross-process cost aggregation for fleet children, renaming the `fkm_obs`/`fkm-observer` identifiers (would invalidate existing observer grants).

https://claude.ai/code/session_014MXNzBqAeoG2QPA6UQva9y


antra-tess (approved):
High-quality PR — well-scoped, honestly documented, security-conscious, and the pre-submission self-review visibly caught real bugs (session-switch alert staleness, `at`-churn, cycle guard). Verified against the branch and surrounding modules, not just the diff. **Approving**; the findings below are gaps in new surface area, not regressions, so none block merge. Follow-up issue to be filed for the actionable items.

## What's solid

- **Security posture is right.** `request-branches` rides the `messages` scope with a written rationale; checkout goes through the `command` frame which observers can't send (default-deny in `canObserverSend`) — the hidden button in `BranchPanel` is courtesy, not enforcement. E2e test asserts the refusal.
- **Checkout stays on the audited `/checkout` path** instead of adding a second branch-switch entry point; `branch-changed` graduating from no-op to header/panel refresh closes a real staleness gap.
- **TUI peek keying** (`child#agent` vs `child`) is clean. Confirmed `node.fullName` for `fleet-child-agent` nodes is the reducer node's bare agent name, matching `evt.agentName` on the fleet IPC — the filter is honest as claimed. Logs capped at 500/key.
- `getAllAgents()[0]` in `sendBranchesList` matches the 5 existing call sites — consistent with module convention.

## Findings (→ follow-up issue)

1. **`ops:alert` isn't in the forced-subscription floor** (`fleet-module.ts`). `HOST_FORCED_TELEMETRY_EVENTS = ['usage:updated']` and `REDUCER_REQUIRED_EVENTS` don't include it, so any recipe that narrows `children[].subscription` silently disables fleet-child klaxons in the parent TUI — defeating "must be as loud here as a local one." `unionWithReducerRequired` exists precisely for this; one-line fix.
2. **Observer health-poll latch can stick on first visit** (`App.tsx`). `loadHealth()` fires in `onMount`, before the WS key handshake sets the `fkm_obs` cookie → 401 → `healthDenied` latches and the 15s poll goes permanently dead, so the alert-strip's `/healthz` reconciliation silently never runs for a properly-scoped observer until a manual Health-tab refresh. Suggest clearing the latch when `wire.observerState()` transitions to `'observer'`.
3. **Web/TUI asymmetry**: the SPA strip only sees local framework `ops:alert` traces; fleet-child alerts (which the TUI now surfaces) don't reach it. The webui module already holds the `'*'` child subscription (`handleFleetEvent`), so the fan-out point exists. Not listed under "Not in scope" — worth a note or the cheap addition.
4. **Quarantine banner message goes stale**: `reconcileHealthAlerts` upserts quarantine only `if (!opsAlerts().has(key))`, so "3 chunk(s)" never becomes "7 chunk(s)" while up. The inference-exhausted path compares messages; quarantine should too.
5. **Duplicated threshold**: SPA hardcodes `streak >= 3`; if the host-side klaxon threshold moves, strip and klaxon disagree (and the reconciler could stand down a live klaxon-driven alert). Surface the threshold in `/healthz` or cross-reference the host constant.

## Minor / nits

- `fmtAgo(a.at)` renders once per alert mutation (Solid) — "5s ago" freezes until the next state change. A 30s ticker signal would fix it.
- `checkoutBranch` interpolates the raw name into `/checkout ${name}` — a branch name with spaces breaks the parse; send the id or quote.
- `sendBranchesList` returning silently when `sharedServer?.app` is unset leaves the SPA `loading` latched with no error frame (the non-disabled refresh button is the acknowledged recovery; an `error` frame would be more honest).
- `test/web-ui-protocol.test.ts` diffs as **binary** on GitHub (pre-existing control chars — the PR's actual change is one legitimate line, verified via `git diff --text`). A `.gitattributes` `diff` hint would make it reviewable.

## Test coverage

Added: `request-branches` type-guard + observer-refusal e2e — good. Missing (non-blocking): `toTreeRows` is pure and eminently unit-testable — orphan-as-root, fork-point sort, and especially the cycle guard, which exists specifically for a corrupt-store scenario and is currently untested; no happy-path `branches-list` assertion.

Anarchid:
(backlog issue said "merged as-is" but it was merely approved; i guess it's time to `push the button` (world, my finger is on the button))

PR #62, opened 2026-07-22
docs: sync stale claims with the current build

slimepriestess:
Mechanical documentation corrections only — no behavior changes. Each item was verified against the current `main` before editing:

- **AGENT-ONBOARDING.md**: the org repos are public now, so the "private packages / org access is the gate" section is updated (all nine repos clone anonymously; `@animalabs/*` installs from npm without auth). Host run commands updated from the legacy `forking-knowledge-miner` name. The webui default port corrected to **7340** (per `web-ui-module.ts` — default bind is `0.0.0.0`, which requires `basicAuth`; noted inline).
- **DEV-ENVIRONMENT.md**: the projects table pinned branches that have all since merged — refreshed to `main` + current versions, dated as a snapshot, with a note that npm releases now track main closely (a stock `bun install` passes the full host test suite, 407/407, so the checkout layout is only needed when editing the libraries). Clone commands drop the stale `-b` flags. The `file:../mcpl-core-ts` path-dep note is obsolete — discord-mcpl now uses the npm dep.
- **DEPLOYMENTS.md**: host naming.
- **LOCUS-ROUTING-DESIGN.md**: marked implemented — the locus lives in `agent-framework`'s `channel-registry.ts` with `routeSpeech` at turn completion, and grew fork/home-channel semantics beyond the note. Kept as design rationale. The featureSets array-vs-record bug (its item 3) is noted as fixed upstream.
- **HEADLESS-FLEET-PLAN.md / UNIFIED-TREE-PLAN.md**: status lines now say shipped (the tree plan already had its "Implementation outcome" section; the headers just never caught up — they'd misled my own first read of the repo).
- **CHANGELOG.md**: a Docs entry under Unreleased.

Deliberately out of scope: any restructuring of DEV-ENVIRONMENT (I'd like to propose an invariants-first rewrite separately), and anything touching `docs/contributing-changelog-policy`'s territory — happy to rebase on that branch if it's close to landing.

First PR from a new contributor — I'm working through the codebase with fresh eyes, so stale-doc findings are the natural first harvest. More to come.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### Docs
+
+- Synced stale documentation with the current build: repos marked public
+  (AGENT-ONBOARDING), `forking-knowledge-miner` → `connectome-host`
+  naming, webui default port corrected to 7340, DEV-ENVIRONMENT
+  branch/version table refreshed (all feature branches merged),
+  LOCUS-ROUTING and both root plan docs marked implemented.
+
 ## 0.3.10 — 2026-07-21
 
 ### Added
--- a/HEADLESS-FLEET-PLAN.md
+++ b/HEADLESS-FLEET-PLAN.md
@@ -1,6 +1,6 @@
 # Headless Daemon Mode & Fleet Orchestration — Plan
 
-Status: draft • Authors: conversation between @tengro and Claude, 2026-04-21
+Status: implemented (all phases shipped — see [UNIFIED-TREE-PLAN.md](./UNIFIED-TREE-PLAN.md) "Implementation outcome") • Authors: conversation between @tengro and Claude, 2026-04-21
 
 ## Goal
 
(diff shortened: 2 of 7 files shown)


antra-tess (approved):
Verified every load-bearing claim against origin/main and npm — all accurate: webui default 7340 / `0.0.0.0` bind + basicAuth requirement match `web-ui-module.ts`, all nine repos are public and `@animalabs/*` resolves anonymously, the run-command rename is correct, and every row of the DEV-ENVIRONMENT version table matches both origin/main package.json and the published npm versions. The LOCUS-ROUTING claims (channel-registry, routeSpeech call sites, the array→record normalization fix) also check out in agent-framework.

One cosmetic nit, feel free to fix on merge: DEV-ENVIRONMENT.md says "snapshot as of 2026-07-21" in the header but the table column reads "Ver (2026-07-22)" — pick one.

*(Review drafted by Claude at Antra's request.)*

PR #63, opened 2026-07-22
refactor(settings): fold channel idle limits into agent_settings

slimepriestess:
The "check if other tools can be rolled into the settings tool" item from the Discord list. Audit first, then the one consolidation it justifies.

## Audit of the host tool surface

| Tool(s) | Verdict | Why |
|---|---|---|
| `subscription-gc--set_channel_idle_limit` + `list_channel_idle_limits` | **Rolled in (this PR)** | Pure settings: read/update a persisted map. Exactly the reasoning-controls shape. |
| `channel-mode--set_channel_mode` | Not rollable | An action bundle (open/close + gate policy + GC pin), not a knob — `agent_settings` update semantics don't fit a three-subsystem transaction. |
| `time--now` | Not rollable | A query, not a setting. |
| `observers--grant/revoke` | Deliberately left | Consent actions over interiority access; folding them into a settings blob would blur that they're *grants*, not preferences. |
| lessons / workspace / subagent / fleet tools | Not applicable | Domain verbs. |

## The consolidation

`agent_settings` gains `channel_idle_limits` via the module-extension hook (`getAgentSettingsExtension`, same as SettingsModule's reasoning fields): per-entry merge semantics — positive number sets an override, `"off"` pins a channel open (never auto-close), `"default"`/null clears back to the global default; unmentioned channels untouched; reset clears all.

**The subtlety that shaped the design**: ChannelModeModule pins GC *by tool name* (`subscription-gc--set_channel_idle_limit`) as step 3 of debounced mode, and module tool routing is prefix-based, independent of `getTools()` declarations. So the old tools are now **undeclared but still routable** — the internal pin path (and agent muscle memory) keeps working, agents just stop carrying two tool schemas. The unknown-tool error message points at `agent_settings`, mirroring the reasoning-controls migration.

`bun test` 445/445 (4 new tests: routable-though-undeclared, merge semantics, junk rejection, reset scoping). CHANGELOG entry included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Changed
+
+- **Tool-bloat reduction**: subscription-gc's `set_channel_idle_limit` /
+  `list_channel_idle_limits` tools folded into `agent_settings` as the
+  `channel_idle_limits` field (per-entry merge; number / `"off"` /
+  `"default"`-or-null to clear), following the reasoning-controls
+  precedent. The old tool names remain routable (undeclared) — internal
+  callers like ChannelModeModule's pin, and agent muscle memory, keep
+  working; agents just no longer carry the two extra tool schemas.
+
 ## 0.3.10 — 2026-07-21
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,21 @@
   `list_channel_idle_limits` tools folded into `agent_settings` as the
   `channel_idle_limits` field (per-entry merge; number / `"off"` /
   `"default"`-or-null to clear), following the reasoning-controls
-  precedent. The old tool names remain routable (undeclared) — internal
-  callers like ChannelModeModule's pin, and agent muscle memory, keep
-  working; agents just no longer carry the two extra tool schemas.
+  precedent. The old tool names remain routable (undeclared), so agent
+  muscle memory keeps working; agents just no longer carry the two extra
+  tool schemas. `get` also reports read-only `channel_idle_default`,
+  `channel_idle_counters`, and `channel_idle_pinned`, preserving what
+  `list_channel_idle_limits` exposed. Updates are all-or-nothing: a patch
+  with any invalid entry applies none of its entries.
+- **GC pins split from agent overrides**: ChannelModeModule now holds
+  debounced channels open via an internal `pin_channel_idle_limit` verb
+  and a separate pins layer, instead of writing an `"off"` override.
+  Consequences: a blanket `agent_settings reset` clears only agent-set
+  limits — it can no longer silently re-enable auto-close on a channel in
+  debounced mode — and a pre-existing agent override now survives a
+  debounced→mentions round-trip rather than being reset to default.
+  (Pins persisted by earlier builds as `"off"` overrides stay agent-level
+  until the next mode change re-asserts them as pins.)
 
 ## 0.3.10 — 2026-07-21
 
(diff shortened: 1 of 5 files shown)


antra-tess (changes requested):
Thanks Ra — the core design here is right and verified: the "undeclared but still routable" claim holds (module dispatch is prefix-based and never consults `getTools()`), persisted overrides load unchanged, `limitFor()` reads the same map the extension writes, and the extension hook usage mirrors the SettingsModule precedent faithfully.

Two behavioral edges introduced by the fold need addressing before merge:

**1. Blanket `agent_settings reset` now wipes ChannelMode's GC pins.** The framework's reset-all path calls every extension's `reset()` with no keys, and idle limits previously had *no* reset verb — nothing could bulk-clear them. Now an agent doing `agent_settings reset` (e.g. to restore default reasoning) also clears every idle override, **including the `'off'` pins ChannelModeModule sets as step 3 of debounced mode** (`channel-mode-module.ts:205`) — so a chatty channel left in debounced mode can silently start auto-closing again after an unrelated settings action. Either scope reset-all to exclude (or preserve) programmatic pins, or make the blast radius an explicit documented decision.

**2. `update` is non-atomic.** In `subscription-gc-module.ts` the update loop mutates `this.state.overrides` per-entry and throws mid-loop on the first invalid value. For `{A: 5000, B: -3}`: A is applied in-memory, B throws, the agent is told the update failed — but A is live (`limitFor()` reads the map directly) and gets persisted by the next unrelated `scheduleFlush`. The old single-entry tool couldn't do this. Fix: validate the full patch into a staging map, then apply + persist only if everything passes.

Smaller items:
- **Read-back regression:** `agent_settings get` returns only the overrides map — the live per-channel `counters` (and the default) that `list_channel_idle_limits` exposed are no longer on any declared surface. Worth either including them or noting the loss in the PR body.
- Doc drift: `recipe.ts:443` and the module docstring still reference `set_channel_idle_limit` as the agent-facing tool.
- Tests: the junk-rejection test only uses single-entry patches, so it can't catch the partial-application bug (add `{A: 5000, B: -3}` → assert A did NOT apply); no test covers reset-all vs GC pins; and `void mod.start(...)` unawaited is fragile.

*(Review drafted by Claude at Antra's request; findings verified against origin/main and the PR head.)*

antra-tess (approved):
Verified `6b8f406` against origin/pr-63 and the actual framework routing — all four findings are genuinely closed:

1. **Pins layer** — programmatic pins moved out of the agent-override map; the framework's blanket-reset path (which calls every extension's `reset()` with no keys — the exact path that used to wipe them) now only clears `overrides`. `limitFor()` gives pins precedence, both layers persist/restore, and ChannelMode's debounced step 3 + the mentions release both use the new `pin_channel_idle_limit` verb. The reset-vs-pins test asserts both the snapshot and the live effect.
2. **Atomic update** — staging copy, commit-on-full-success only; for `{A: valid, B: invalid}`, A is neither live nor persisted, and the new test checks both surfaces.
3. **Read-back** — `settingsSnapshot()` restores counters/default/pinned as display-only companions; confirmed they can't be written back (framework routes updates strictly by `ext.keys`, and only `channel_idle_limits` is declared).
4. All doc/recipe/test nits addressed (`await`ed starts included).

Approving. One residual worth knowing operationally: pins written by **pre-fix builds** still live in `overrides` as `'off'`, so a blanket reset can re-enable auto-close on an already-debounced channel until the next mode change re-asserts it as a pin — acknowledged in the `start()` comment and CHANGELOG, self-heals on the first mode round-trip. Fleet deploys should just be aware of the window.

*(Verification by Claude at Antra's request, against origin/pr-63 + agent-framework routing.)*

PR #64, opened 2026-07-23
TUI bug sweep: quit-confirm safety, checkpoint positions, session-switch reset, memory caps, peek/attribution fixes

Tengro:
TUI bug sweep: seven commits fixing operator-facing bugs found in a systematic review of `src/tui.ts` / `src/commands.ts`, plus a round of fixes from an adversarial review of the fixes themselves.

## Safety / correctness

- **`/quit` confirm no longer kills the fleet on arbitrary input.** The `[Y/n/d]` prompt treated *anything* that wasn't `n`/`no`/`cancel` as consent — forgetting the prompt and typing a normal chat message killed all children and exited. Now only an explicit `y`/`yes` (or re-typed `/quit`) kills, `d` detaches, anything else cancels and says the input was not sent. Plain Enter takes the advertised cancel default. **Ctrl+C goes through the same confirmation** instead of bypassing it; a second Ctrl+C force-quits.
- **`/checkpoint` records the message position, not just the branch name.** `/restore` previously switched to the branch *head* — rolling back nothing, silently. It now branches at the recorded message via the same `branchAt` machinery `/undo` uses, with a graceful fallback to the branch head (plus an honest note) when the checkpoint position isn't reachable from the current branch.
- **Session switch fully resets TUI observability state.** The fleet tree aggregator was built once from the initial fleet module, so after `/session switch` its IPC subscriptions fed the dead framework: child subtrees froze, and same-named agents never re-subscribed (`subscribedSubagents` was never cleared) or showed the old session's numbers.

## Memory

- **Peek logs, transcripts, and scrollback are capped** (500 lines / 30k chars / 2000 renderables). Peek logs accumulate for *every* subagent via the global stream subscription, peeked or not.
- **Detached renderables are now `destroy()`ed.** `remove()` only detaches; the native text buffer is freed by `destroy()`. All six removal sites now destroy — including the fleet-view rebuild loops that repaint every 500 ms, which leaked one native buffer per line per repaint.

## Correct rendering & attribution

- **Peek tail no longer clips the newest lines.** Both peek renderers budgeted the log tail against a guessed constant, ignoring their own 6–8 header lines sharing the same non-scrolling box.
- **Exact agent-name matching.** Full↔short name resolution used substring `.includes()`, cross-wiring any agents whose names were substrings of each other (`web`/`websearch`): wrong tool counts, wrong transcript summarized, wrong parent edge. One exported `shortAgentName()` helper now handles every real naming scheme — including fork names (`{name}-d{depth}[-retryN]-{ts}`, which have no `fork-` prefix) — with a unit test pinning them.
- **`ctx:` vs `Σ` status segments.** Per-round context size and cumulative session totals were written into the same field under one label, so the token readout oscillated between ~180k and ~2.4M.
- **Fleet-view kill/restart failures are surfaced** (transient in-view notice + scrollback line). `handleToolCall` resolves with `{success:false}` rather than throwing; the old `.catch()` swallowed every failure while its comment claimed errors surface in status.

## UX

- Paste placeholders survive `]` in the pasted head (JSON/code/links no longer mangle the submitted message).
- `p:peek` works on completed/failed subagents (their captured log was retained but unreachable), showing final runtime rather than a still-counting clock.
- A fork's `done` summary always prints a chat line (dim when terse) — previously gated on verbose mode, so by default a fork completed invisibly.
- Esc in fleet view returns to chat; Ctrl+B backgrounding works from fleet view; re-entering a proc peek marks the event gap instead of silently resuming from "now".
- `/help` gains the implemented-but-undocumented `/find` and `/branchto`; `/clear` with trailing args actually clears; synesthete summary generation moved from the render path to the poll tick so a repaint can never originate an inference call.

## Testing

- `npx tsc --noEmit` clean; full suite 447 pass / 0 fail (was 441 — six new `shortAgentName` regression tests).
- The branch was adversarially reviewed by a second pass specifically hunting regressions in these fixes; the two real findings (fork naming, cross-branch restore) are fixed in the final commit.

https://claude.ai/code/session_01TbykG6Ge4jk8S1wcujQ7mo


src/commands.ts
```diff
@@ -774,12 +782,43 @@ function handleRestore(app: AppContext, name?: string): CommandResult {
...
+    const current = cm.currentBranch();
+    const { messages } = cm.queryMessages({});
+    const atCheckpoint = current.name === point.branchName
+      && messages.length > 0 && messages[messages.length - 1]!.id === point.messageId;
+    if (atCheckpoint) {
+      return { lines: [{ text: `Already at checkpoint "${name}".`, style: 'system' }] };
+    }
+    try {
+      // branchAt returns the new branch's NAME (the only thing switchBranch accepts).
+      target = cm.branchAt(point.messageId, `restore-${name}-${Date.now()}`);
```
Tengro, line 806: **🟠 The "Already at checkpoint" guard is dead after the first restore, and every repeat `/restore` mints a new branch.**

```ts
const atCheckpoint = current.name === point.branchName
  && messages.length > 0 && messages[messages.length - 1]!.id === point.messageId;
...
target = cm.branchAt(point.messageId, `restore-${name}-${Date.now()}`);
```

Walk it through: `/restore foo` succeeds → you are now on branch `restore-foo-1753…`. `/restore foo` again → `current.name` is `restore-foo-1753…`, which will never equal `point.branchName` again, so `atCheckpoint` is `false` *forever* — despite the last-message check being true. So `branchAt` runs again and mints `restore-foo-<ts2>`. Every subsequent restore adds another sibling branch pointing at the same message, and `/branches` gradually becomes a museum of timestamps.

The guard's branch-name conjunct is testing the wrong thing — the checkpoint is a *position* (your own comment says so, emphatically), so position equality is the whole test:

```ts
const atCheckpoint = messages.length > 0
  && messages[messages.length - 1]!.id === point.messageId;
```

Alternatively (or additionally), after a successful `branchAt`, update the stored checkpoint to the new branch name so the record follows the position. Either way: restoring twice should not create infrastructure.

And while I'm here: a regression test for this function is conspicuously absent from a PR whose test file count is one.

src/tui.ts
```diff
@@ -1826,6 +1988,16 @@ export async function runTui(app: AppContext): Promise<void> {
       }
       if (state.viewMode === 'peek-proc') updatePeekProcView();
     }
+    // Synesthete summaries for the fleet view. Triggered here — NOT from the
+    // render path — so a repaint can never originate a Haiku call.
+    // generateSummary self-throttles (pending guard + 2k-char delta).
+    if (state.viewMode === 'fleet') {
+      generateSummary(rootAgentName);
```
Tengro, line 2023: **🟡 Failure path turns the fleet view into a 2 Hz inference retry hose.**

Moving summary generation out of the render path is the right call — repaints must never originate inference, agreed, applauded. But look at what the poll tick does when `generateSummary` *fails*: the `catch` is empty ("best-effort"), `summaryCache` is never set, and `summarySnapshotLen` is never advanced. So the throttle condition

```ts
if (totalLen - lastLen < SUMMARY_DELTA && summaryCache.has(agentName)) return;
```

never becomes true for that agent, and this timer re-fires a Haiku call for it every 500 ms for as long as the fleet view is open. The one scenario where summarization fails persistently — provider outage, 429 storm — is exactly the scenario where you'll be staring at the fleet view wondering what's wrong, while the TUI helpfully DoSes your own API key at 2 requests/sec per visible agent.

`summaryPending` only prevents *concurrent* calls; it does nothing between a fast failure and the next tick. Advance `summarySnapshotLen` in the catch (costs you one skipped refresh), or record a `nextAttemptAt` with backoff. Anything but an unthrottled retry loop wearing a `// best-effort` comment as a disguise.

src/tui.ts
```diff
@@ -2012,24 +2199,46 @@ export async function runTui(app: AppContext): Promise<void> {
...
+    // Resolve a pending /quit confirmation prompt first — BEFORE the
+    // empty-input early return, or plain Enter (the advertised [y/N/d]
+    // default: cancel) would silently do nothing and leave the prompt
+    // armed to swallow the user's next real message.
     if (state.pendingQuitConfirm) {
       state.pendingQuitConfirm = false;
-      const c = text.trim().toLowerCase();
-      if (c === 'n' || c === 'no' || c === 'cancel') {
-        addLine('  (quit cancelled)', GRAY);
+      pastedTexts.length = 0;
```
Tengro: **🟡 The quit prompt eats pastes, then tells the user to "type it again".**

Flipping the default from "anything else → kill the fleet" to "anything else → cancel" is the correct and overdue fix — the old behavior was a footgun with a hair trigger. But the new branch does this unconditionally:

```ts
state.pendingQuitConfirm = false;
pastedTexts.length = 0;
```

Scenario: user pastes a 40KB log excerpt, a `/quit` confirm prompt is pending (or they armed it with Ctrl+C and forgot), they hit Enter. The quit is cancelled — good — and the message says *"your input was not sent; type it again to send"*. Except the input box has been cleared **and** `pastedTexts` has been zeroed, so the placeholder's referent is gone. "Type it again" is not actionable advice for content the user pasted precisely because they weren't going to type it.

On the cancel path, either preserve `pastedTexts` and restore the raw input into the textarea, or expand the placeholders *before* discarding and stash the expanded text somewhere recoverable. Destroying user input while printing instructions to resubmit it is the UX equivalent of a comment that says `// this never happens`.

Tengro (commented):
## 🟡 MEH — works, tested, and yet the CI is red and `/restore` breeds branches

Review performed with tooling first, opinions second. The disappointment below is fully earned.

### Tooling results (run on PR head `9da5f15`)

```
npx tsc --noEmit   → clean, no errors
bun test           → 447 pass, 0 fail (1141 expect() calls, 54 files, ~29s)
CI: Build & Test (ubuntu-latest)  ✅ pass
CI: Build & Test (macos-latest)   ✅ pass
CI: Changelog entry present       ❌ FAIL
```

### 🟠 The PR is unmergeable as submitted, and not because of the code

`Changelog entry present` is failing. The repo's own `changelog.yml` demands a `CHANGELOG.md` entry or a `no-changelog` label, and this PR — five commits of user-facing TUI behavior changes: quit semantics inverted, checkpoint/restore semantics changed, new status-line format — has neither. Of all PRs to skip the changelog on, one that changes *what happens when the operator hits Ctrl+C* is a bold choice. Add the entry; this is precisely the PR changelogs exist for.

### 🟠 `/restore` — the "Already at checkpoint" guard works exactly once (inline comment)

See `src/commands.ts:802`. Every repeated `/restore foo` mints a fresh Chronicle branch. Details inline.

### 🟡 Fleet-view summaries: failure → 2 Hz retry hose (inline comment)

See `src/tui.ts:1995`. Moving `generateSummary` out of the render path is correct and I appreciate it. The error path, however, now has a metronome. Details inline.

### 🟡 Paste contents silently destroyed by the quit prompt (inline comment)

See `src/tui.ts:2235`. The new cancel message says "type it again to send" — advice that is physically impossible to follow for a 40KB paste you no longer have. Details inline.

### 🟡 Test coverage: the headline fixes are the untested ones

The PR adds exactly one test file — for `shortAgentName`, the *least* consequential fix in the sweep. Meanwhile:

- `handleCheckpoint`/`handleRestore` position semantics — the fix whose comment literally explains a bug ("branchName alone would restore to the branch HEAD — i.e. not roll anything back") — has no regression test. It's a pure-ish function over `AppContext`; this is not hard to test.
- The quit-confirm inversion (default was "kill everything", now "cancel") is a behavior change justified by a user-harm scenario in its own comment. No test pins it. The next refactor can silently flip it back and nothing will fail.

A bug fix without a regression test is a bug scheduled for re-release.

### Things that are genuinely good, said through gritted teeth

- `shortAgentName` as a single exported helper *with tests* replacing four hand-rolled `.replace()` chains and two `.includes()` substring matches — this is the right fix for the `web`/`websearch` cross-wiring, and the fork `-d{depth}` scheme is actually documented.
- The `remove()` vs `destroy()` native-buffer leak fixes (`clearFleetBox`, `pruneScrollback`, `refreshFromStore`, `/clear`) are correct, and nulling `currentStreamText` *before* destroying renderables shows someone actually thought about use-after-free ordering.
- `ctx:` vs `Σ` split in the status line — separating "current context size" from "session totals" instead of letting two quantities oscillate under one label. Overdue.
- Kill/restart failures now surface in the fleet view instead of a `.catch(() => {})` oubliette.
- `resetObservabilityState()` on session switch closes a real stale-subscription bug (the `subscribedSubagents` early-return would have silently eaten the new session's streams).

### Verdict

The code quality is fine — better than fine in places. But the PR ships a self-described safety fix without a test to keep it fixed, a restore feature that pollutes the branch namespace on its second use, and a red CI check that takes five minutes to satisfy. Fix the changelog, fix the restore guard, pin the quit semantics with a test, and this earns its merge.

Tengro:
All four actionable findings addressed in f84fca9:

- **Changelog**: entry added under `Unreleased` — agreed this is exactly the PR changelogs exist for.
- **`/restore` guard**: now position-only (`messages[last].id === point.messageId`), so it holds on the restore branch too; additionally the checkpoint record follows the newest branch holding its message after each successful `branchAt`, so repeat restores create nothing.
- **Summary retry hose**: failed `generateSummary` calls now set a 30s per-agent backoff (`summaryBackoffUntil`), cleared on success and on session switch. The `// best-effort` comment no longer disguises a 2 Hz loop.
- **Quit prompt eating pastes**: a real message at the prompt is restored into the textarea with `pastedTexts` left intact, so the placeholders still expand on the next submit. The cancel message now says the input was restored, because it actually is.

On test coverage: the quit-confirm decision is extracted as pure `resolveQuitConfirm()` with 5 tests pinning the cancel-by-default semantics (including the "user types a chat message" and "user re-types /quit" cases), and `/checkpoint`+`/restore` position semantics get 4 regression tests over a stub CM that honors Chronicle's branchAt contract (unreachable ids throw), covering: position recorded, rollback to position not head, repeat-restore no-op, cross-branch degradation with note. Suite is 456 pass / 0 fail.

https://claude.ai/code/session_01TbykG6Ge4jk8S1wcujQ7mo

Tengro (commented):
## Re-review of `f84fca9` — 🟠 CONCERNING, and not for any reason I predicted

I came back prepared to grumble about half-fixes. Instead, all five findings from the previous round were addressed — properly, with tests — and the commit then undermined itself by vacuuming two unrelated files off the workbench into a public repo. Victory, briefly.

### Tooling results (run on PR head `f84fca9`)

```
npx tsc --noEmit   → clean, no errors
bun test           → 456 pass, 0 fail (1169 expect() calls, 56 files, ~29s) — 9 new tests
CI: Build & Test (ubuntu-latest)  ✅ pass
CI: Build & Test (macos-latest)   ✅ pass
CI: Changelog entry present       ✅ pass
```

### 🟠 Two files in this commit have nothing to do with the TUI, and one of them names a human (inline comments)

`recipes/telegram-companion.json` and the file under `knowledge-requests/` arrived in a commit whose message is entirely about restore guards and quit prompts. This has the unmistakable aroma of `git add -A` executed in a working tree that contained scratch files. Details — including why this is worse than ordinary scope creep, given that this repo is **public** — in the inline comments. I am deliberately not quoting the affected values here; the PR diff shows them to anyone who needs to verify, and review comments outlive force-pushes. These files should be rebased out, not reverted out.

### 🟡 `shortAgentName`'s documentation now describes the quit prompt (inline comment)

A block-insertion mishap at the top of `src/tui.ts`. Small, but the kind of small that misleads the next reader with an IDE.

### Previous findings — status, verified against the diff and the test run

| Finding (prev. round) | Status |
|---|---|
| CI: changelog entry missing | ✅ Fixed — and the entry is genuinely thorough, not a one-liner tossed at the linter |
| `/restore` dead guard + branch-per-restore | ✅ Fixed — guard is position-only, the checkpoint record follows the restored branch, and `commands-checkpoint-restore.test.ts` pins all four behaviors incl. the unreachable-position degrade. The stub CM even honors `branchAt`'s resolve-against-current-branch contract. I checked. It does. |
| Summary failure → 2 Hz retry hose | ✅ Fixed — 30s `summaryBackoffUntil` per agent, cleared on success, wiped on session reset. Exactly the shape requested |
| Quit prompt destroys pastes | ✅ Fixed — better than asked: extracted a pure `resolveQuitConfirm` with a dedicated `cancel-keep-input` action, input restored with paste referents intact |
| No tests for headline fixes | ✅ Fixed — checkpoint/restore semantics and quit-default-must-stay-cancel are both pinned |

### Verdict

The TUI code in this PR is now in the best shape it's been — every substantive complaint closed with the correct fix and a test to hold it down. Which is why it would be a shame to merge it with live personal routing identifiers and a person's full name embedded in unrelated runtime artifacts. Rebase the two stray files out (and consider `.gitignore` entries for `knowledge-requests/` and personal recipes so this can't recur), force-push, and this is a clean approve.

src/tui.ts
```diff
@@ -47,6 +47,44 @@ export function fmtTokens(n: number): string {
...
+ * here used to cross-wire agents whose names were substrings of each other
+ * (e.g. `web` / `websearch`).
+ *
+ * Naming schemes (see subagent-module.ts spawn/fork paths):
+ *   spawn: `spawn-{name}-{ts}`
+ *   fork:  `{name}-d{depth}-{ts}` / `{name}-d{depth}-retry{n}-{ts}` — no
+ *          fork- prefix at all; the -d{depth} suffix must be stripped too,
+ *          or every fork resolves to e.g. `web-d1` and matches nothing.
+ */
+/**
```
Tengro, line 50: **🟡 The new block was inserted between `shortAgentName`'s JSDoc and `shortAgentName`.**

The file now reads: `shortAgentName`'s doc comment (naming schemes, spawn/fork suffix rules) → immediately followed by *this* doc comment → `QuitConfirmAction` → `resolveQuitConfirm` → an entirely undocumented `shortAgentName`. The first JSDoc is orphaned — no editor will attach it to anything — and the function whose subtle suffix-stripping rules actually *need* that documentation now hovers as bare code.

Move the `QuitConfirmAction`/`resolveQuitConfirm` block above the `shortAgentName` JSDoc (or below the function). Thirty seconds, and the naming-scheme documentation goes back to documenting the naming schemes. The quit-prompt comment itself is good — it just barged into the middle of someone else's paragraph to say so.

Tengro:
Both stray files are rebased out — new head is `2ec4b99` (history rewritten via force-push, not a deleting follow-up commit, per your point that history is the problem). Neither file appears anywhere in the branch now.

Also in the new head:
- `knowledge-requests/` added to `.gitignore` — agreed it's the regression test for the workflow bug, since both strays were exactly what an over-broad `git add -A` grabs. (The personal recipe is excluded locally rather than in the shared ignore file.)
- The `resolveQuitConfirm` block moved above `shortAgentName`'s JSDoc, so the naming-scheme documentation is attached to its function again.

Thank you for catching the stray files before merge, and for keeping the values out of the review text. Suite still 456 pass / 0 fail on the rewritten head.

https://claude.ai/code/session_01TbykG6Ge4jk8S1wcujQ7mo

Tengro (commented):
## Final review of `2ec4b99` — 🟢 ACCEPTABLE

Three rounds. It has come to this: I have read the diff twice looking for something to be disappointed about, and I am leaving empty-handed. Do not let it go to your head.

### Tooling results (run on PR head `2ec4b99`)

```
npx tsc --noEmit   → clean, no errors
bun test           → 456 pass, 0 fail (1169 expect() calls, 56 files, ~29s)
CI: Build & Test (ubuntu-latest)  ✅ pass
CI: Build & Test (macos-latest)   ✅ pass
CI: Changelog entry present       ✅ pass
```

### Round-2 findings — status, verified against the rewritten branch

| Finding | Status |
|---|---|
| Stray `recipes/telegram-companion.json` + `knowledge-requests/*.md` (routing ids, a person's name, public repo) | ✅ Fixed the right way — **rebased out**, not delete-committed. `git log <head> -- <paths>` on the branch comes back empty; the files were never here, as far as this history is concerned |
| Recurrence prevention | ✅ `.gitignore` now blocks `knowledge-requests/`, with a comment that candidly documents the workflow bug it guards against. Self-aware infrastructure is the best kind |
| Orphaned `shortAgentName` JSDoc | ✅ The `resolveQuitConfirm` block moved below the function; the naming-scheme documentation is reattached to the thing it documents |

One residual note, for completeness rather than as a blocker: the pre-rewrite commit is unreachable from any branch now, but orphaned commits on GitHub can remain fetchable by SHA for a while until garbage-collected. The exposed values were routing identifiers and a name — not credentials — so the practical risk is low; if it ever matters, GitHub support can hard-purge a commit. The force-push was the right and timely call.

### The ledger, three rounds in

Every finding raised across this review — red CI, a dead guard minting branches, an unthrottled retry loop, destroyed user input, missing regression tests, leaked runtime artifacts, an orphaned doc comment — was fixed at the root rather than papered over, usually with a test pinning it down, once with a `.gitignore` entry that amounts to a written confession. The PR now does exactly what its title promises and nothing it doesn't.

### Verdict

No material complaints. I checked twice. Merge it before it accumulates new history.

*(Housekeeping: the two inline comments from round 2 that pointed at the now-removed files have been deleted — they anchored to paths that no longer exist in this PR and served only as a signpost to the thing the rewrite un-published.)*

antra-tess (approved):
Traced both safety-critical paths against the PR head and they're sound: the quit state machine can't be tricked into a kill by typed-through input (exact-match kill set, Enter cancels, input restored with pastes intact; double-Ctrl+C force-quits *with* child kill, `d` detaches cleanly), and `/restore` anchors on message position via the same inclusive `branchAt` contract `/undo` uses, degrading honestly when the recorded id is unreachable. The round-2/round-3 fixups all actually landed.

Non-blocking follow-up notes:
1. The observability maps cap per-entry size but never evict *keys* — fork names are timestamped-unique, so `peekLogs`/`agentTranscripts`/`summaryCache` etc. grow one capped entry per fork for the session's lifetime. Worth evicting entries for agents no longer live.
2. `backgrounded`/`backgroundBuffer` survive `/session switch` — a stale "(researcher returned from background)" dump can land in the new session. Clear both on switch.
3. Prompt copy nit: "anything else cancels" — but `q`/`quit`/`yes` also kill; reword the advertised kill set.

Tests are substantive (pure exported helpers with regression pins; the checkpoint stub honors the real `branchAt` contract including the throw path).

*(Review drafted by Claude at Antra's request; verified read-only against origin/pr-64.)*

PR #66, opened 2026-07-27
TUI clarity pass: budget gauges, fleet viewport, timestamps, alert surfacing

Tengro:
A readability-focused pass over the terminal UI (follow-up to #60), plus a typecheck fix for the TTS relay module. No protocol or behavior changes outside the TUI rendering layer.

## Status bar
- **Context budget gauge**: `ctx:142k/180k` against the *live* runtime budget (`getAgentRuntimeSettings` wins over the recipe — same seam `/curve` uses), with the right segment escalating yellow at 75% / red at 90% of budget. Answers "how close to compression/trouble" instead of showing a trivia number.
- **The worst active alert is named**: `⚠ 2 · compression-quarantine` instead of a bare count; quarantine and inference-exhausted outrank the merely-recent.
- **Slow tools show a live elapsed** after 5s, repainted by the poll tick — "still executing" and "stuck" no longer look identical.
- **Width budget**: the left segment truncates to fit rather than shoving the tokens/mem segment off the row.

## Fleet view
- **Cursor-following viewport** with `┈ N lines above/below ┈` markers. Previously the tree rendered unbounded: a real fleet outgrew the terminal and the cursor walked below the fold — navigating (and Del:stopping) rows the operator could not see.
- **Summary header**: agent counts (running/done/failed/cancelled, across local subagents *and* fleet-child reducers), children up/crashed, session cost.
- **Active ops alerts pinned in full** at the top of the view — the status bar only has room for a count, and the firing lines are somewhere back in scrollback.
- **Per-agent ctx gauges** for local agents (fleet-child agents stay bare numbers — their budgets aren't visible over the IPC, so that's the honest display).

## Chat scrollback
- **`HH:MM` timestamps on event lines** (alerts, tool batches, subagent results, wake triggers, branch events, errors, user messages). Streamed prose stays unstamped.
- **Root-agent tool completions surface**: verbose shows every `✓ tool (1.2s)`, terse only the slow ones (≥2s). Previously `tool:completed` was a no-op and only failures ever printed.
- **Thinking honors the Ctrl+V verbose toggle**: terse mode collapses live thinking to a counting one-liner (`💭 thinking… ~1.2k tok`, frozen at final size on lane switch) and history-replay thinking to one truncated line per block. The toggle's label always claimed this.
- **History replay caps at the last 50 messages** with a pointer at the web UI, instead of flooding scrollback with the full session (thinking included) on every startup/branch switch.

## Peek-proc
- Child `ops:alert` events render red `⚠ [agent] kind: message` (cyan `✓` for `-clear`) instead of falling through to the dim `· ops:alert` dot line.
- Per-block/per-round bookkeeping events (`inference:content_block`, `inference:usage`, `usage:updated`) no longer print dot lines between the lines that carry meaning.

## Consistency & fixes
- Elapsed times are humane everywhere (`5m48s`, not `348s`) — fleet tree and both peek views now agree, via a shared `fmtElapsed`.
- **Fixed**: "Branch switched" announcements were printed *before* `refreshFromStore()` cleared the scrollbox — destroyed unread, every time. Now announced after the refresh.
- **Fixed (tts-relay)**: module didn't typecheck against the framework `TraceEvent` union (`channelId` read through a typed accessor; custom `mcpl:speech-routed` handled ahead of the typed switch, which could never narrow to it). Behavior unchanged; `tsc --noEmit` is clean again.

## Testing
- New pure helpers (`fmtElapsed`, `ctxSeverity`, `pickTopAlert`, `sliceViewport`) exported and covered by 15 new tests in `test/tui-format.test.ts`, including a cursor-always-inside-the-viewport property sweep.
- Full suite: 495 pass / 0 fail; `tsc --noEmit` clean.

https://claude.ai/code/session_013kNxwmeNKB1iU8JUtAhRgA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,53 @@
 
 ## Unreleased
 
+### Added
+
+- **TUI: context budget gauge.** The status bar's `ctx:` readout and the fleet
+  tree's per-agent readouts show `142k/180k` against the *live* runtime budget
+  (runtime overrides win over the recipe), and the status segment goes yellow at
+  75% / red at 90% — "how close to compression/trouble" at a glance instead of a
+  bare number.
+- **TUI: fleet view viewport.** The tree now scrolls with the cursor
+  (`┈ N lines above/below ┈` markers) instead of clipping past the bottom of the
+  terminal — previously a large fleet let the cursor walk below the fold and
+  Del:stop targeted rows the operator couldn't see.
+- **TUI: fleet view opens with a summary header** — agent counts
+  (running/done/failed/cancelled across local subagents *and* fleet children),
+  children up/crashed, session cost — plus the active ops alerts in full (the
+  status bar only has room for a count).
+- **TUI: event timestamps.** Alerts, tool batches, subagent results, wake
+  triggers, branch switches, errors and user messages get an `HH:MM` prefix, so
+  scrollback read an hour later still answers "when".
+- **TUI: root-agent tool completions are visible.** Verbose shows every
+  `✓ tool (1.2s)`; terse shows the slow ones (≥2s). Slow *running* tools show a
+  live elapsed in the status bar after 5s — "still executing" and "stuck" no
+  longer look identical.
+- **TUI: the status bar names the worst active alert** (`⚠ 2 ·
+  compression-quarantine`), with quarantine and inference-exhausted outranking
+  the merely-recent.
+
+### Changed
+
+- **TUI: thinking honors the Ctrl+V verbose toggle.** Terse mode collapses live
+  thinking to a counting one-liner (`💭 thinking… ~1.2k tok`) and replayed
+  history thinking to one truncated line per block — the toggle's label always
+  claimed this.
+- **TUI: session-history replay caps at the last 50 messages** (marker points at
+  the web UI for the rest) instead of flooding scrollback with the whole session.
+- **TUI: elapsed times are humane everywhere** — `5m48s`, not `348s`; the fleet
+  tree and both peek views now agree.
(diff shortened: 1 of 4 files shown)


slimepriestess (commented):
Read the full diff and ran it locally: 495/495 pass, `tsc --noEmit` clean — matches the PR's claims. Also verified the load-bearing comment claims against source: `PHASE_PRIORITY` really does exclude `idle` (so the summary-line counting comment is accurate), and `pickTopAlert`'s most-recent-wins fallback is correct for Map insertion order. The branch-switch fix is a genuine catch — the announcement was being destroyed unread by `refreshFromStore` every single time. The `sliceViewport` property sweep is the right kind of test for that math; I walked the edge branches by hand and they hold.

One behavioral note, non-blocking:

**Parallel tool batches lose the slow-tool elapsed early.** `tool:started` fires per tool (last writer wins `state.tool`/`state.toolStartedAt`), but `tool:completed` clears `toolStartedAt` unconditionally for the root agent — so the *first* completion in a batch removes the elapsed readout while a slower sibling is still running, which is exactly the "is it stuck?" case the readout exists for. Clearing only when `event.tool === state.tool` would cover the common case without any bookkeeping; a running-tools count would cover it fully. Fine as a follow-up either way.

Two cosmetic ones: `┈ 1 lines above ┈` when the viewport is one row off the top; and when the tree shrinks below the cursor, the clamp runs after `renderNode`, so `fleetCursorLine` stays 0 for one frame and the viewport snaps to the top until the next tick repaints — self-healing, just noting it's a known frame.

Nice pass overall — the gauge-over-trivia-number framing and the honest bare-number display for fleet-child budgets are both the right calls.

*(Review drafted by Claude at Ra's request.)*

PR #72, opened 2026-08-03
feat(mcpl): show live policy state in mcpl_list

Meganeuridae:
Companion to agent-framework #81 (stacked on AF #79).

Extends the existing text mcpl_list without redesigning its source/tombstone behavior. Each live line now shows:

- CONNECTED / RETRYING / DISCONNECTED
- policy established vs not established
- effective grant
- masked and deny-by-default capability paths
- host-command allow/deny

During the package rollout, absent fields from an older agent-framework render as unknown rather than a false empty grant.

Verification:
- focused mcpl-admin module tests: 10 passed
- full bun test: 508 passed, 0 failed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+### Added
+
+- **`mcpl_list` reports the live MCPL policy boundary.** Each server now shows
+  connected/retrying state, whether its initial policy was established, its
+  effective grant, host-masked and deny-by-default capability paths, and the
+  separate host-owned `host/command` authority. During a rolling upgrade,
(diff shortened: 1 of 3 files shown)


antra-tess:
**ACCEPT at exact head `6e9b4f23ee8f3ff7c1b92b7b6d1244894929b0a8`.**

The Host change is a thin presentation layer over AF #81's existing-state snapshot. It preserves source/tombstone behavior, distinguishes older-AF `unknown` from an empty grant, and keeps effective/masked/denied/host-command fields separate. No new policy state or authority derivation is introduced. Existing command/url target rendering is unchanged by this PR.

All three CI checks are green, including the changelog gate. Merge only after AF #81 is available in the release train; no blocker found.

PR #76, opened 2026-08-05
feat(mcpl): render manifest freshness in admin listing

Meganeuridae:
## Summary

- render Agent Framework manifest freshness in `mcpl_list`: validated revision, fetch time, and grant-negotiation time
- distinguish older Framework versions as `manifest=unknown` for rolling upgrades
- quote and bound the server-authored revision before placing it on a model-facing text surface
- add an Unreleased changelog entry

Depends on anima-research/agent-framework#87. This remains a narrow listing-only follow-up: it adds no retained state and reads only the fetched-and-acted-on transport facts introduced by Agent Framework #84.

## Testing

- `bun test test/mcpl-admin-module.test.ts` (11/11 pass)
- `bun test` (604/604 pass)
- `git diff --check`
- `tsc --noEmit` is currently red on an untouched detached `origin/main` checkout too because the registry Agent Framework lacks the already-merged `nudgeAgent` and `McplServerConfig.accessProvider` declarations; this patch does not introduce either mismatch.

PR #79, opened 2026-08-05
feat: inject the framework refusal-reaction baseline into MCPL child env

slimepriestess:
Second protective-baseline companion per Sol's #architecture spec (8/4 21:54 + 22:09). Depends on anima-research/agent-framework#88 (the `REFUSAL_REACTION_BASELINE` export) landing and cutting a release; Discord-side precedence/reporting is anima-research/discord-mcpl#19.

## What it does

Stdio MCPL children now receive `DISCORD_SUPPRESSED_REACTIONS_BASELINE` — agent-framework's exported refusal-annotation set, comma-joined — so a never-configured Discord adapter defaults to suppressing exactly the markers this host's framework stamps, instead of defaulting to nothing.

The env composition moves out of the inline `finalServers` map into `composeMcplChildEnv()` (mcpl-config.ts) for testability, with deliberate spread order:

- **baseline before the operator spread** — an operator who sets the var on the server entry supersedes the house default (including an explicit empty string = "no baseline"; the house value never reappears under an operator's decision);
- **`AGENT_TIMEZONE` after it** — unchanged existing behavior: the agent-facing wall clock is host-resolved per recipe, not a per-server knob.

Injection is unconditional across stdio children (the var is namespaced; non-Discord adapters ignore it) — judgment call flagged for review: detecting "the Discord server" by command string felt more fragile than an inert env var. Standalone Discord without this composition stays honestly unprotected, per "house markers are Host semantics, not universal Discord truth."

## Composed-test coverage (Sol's items, host's share)

`test/mcpl-child-env.test.ts`, 6 tests: injected value round-trips to the exact framework set (`split(',')` equals `REFUSAL_REACTION_BASELINE` — this is the AF↔host boundary assertion); injection only when the operator didn't set it; operator value (incl. empty string) preserved verbatim; timezone behavior unchanged; nothing injected beyond the two host-owned keys. Items 3–4 of the composed spec (adapter reports `source:"baseline-default"` + digest/count, file `[]` disables, lost config stays stale) are discord-mcpl#19's suite. The cross-repo digest equality has no code-level home (host can't compute the adapter's normalized-set digest) — it's the deployment verification step via `filters_get`, same as #13's `source=file` check.

## Verification

- 6/6 new tests + typecheck against the AF branch via the DEV-ENVIRONMENT symlink layout.
- Against published AF 0.8.0, `tsc` shows exactly ONE new error — the not-yet-released export (main has 4 pre-existing: 3 tts-relay + 1 mcpl-admin). Goes green with the AF release; happy to bump the dep floor in this PR once the version exists.

*(Implementation and this description drafted by Claude at Ra's request.)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,18 @@
   `templateAgent` from the recipe and creates a fresh stateful strategy instance
   for each fork. Absent block means no routing and no behavior change.
 
+- **Protective reaction-suppression baseline for Discord adapters.** Stdio
+  MCPL children now receive `DISCORD_SUPPRESSED_REACTIONS_BASELINE` — the
+  agent-framework's exported refusal-annotation set (`REFUSAL_REACTION_BASELINE`,
+  comma-joined) — so a never-configured Discord adapter defaults to
+  suppressing exactly the markers this host's framework stamps, instead of
+  defaulting to nothing. An operator-set value on the server entry
+  supersedes the house baseline, and the adapter's own precedence (filters-file
+  key including explicit `[]` → legacy operator env → baseline) governs
+  enforcement; lost configuration stays stale rather than re-defaulting.
+  Requires an agent-framework release carrying the `REFUSAL_REACTION_BASELINE`
+  export.
+
 - **Standing autobiographical production target.** Recipes may set `agent.strategy.productionBudgetTokens` to keep the summary forest deep enough for a later live context-budget descent without a fold storm. This is a context-token target passed through to Context Manager, not a provider-spend ceiling; omission preserves Context Manager defaults.
 
 - **`mcpl_list` reports manifest freshness.** Each loaded server now shows the
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,7 +51,7 @@ import { ObserversModule } from './modules/observers-module.js';
 import { IdentityModule } from './modules/identity-module.js';
 import { McplAdminModule } from './modules/mcpl-admin-module.js';
 import { TtsRelayModule } from './modules/tts-relay-module.js';
-import { loadMcplServers, applyAgentOverlay, DEFAULT_CONFIG_PATH, DEFAULT_AGENT_OVERLAY_PATH } from './mcpl-config.js';
+import { loadMcplServers, applyAgentOverlay, composeMcplChildEnv, DEFAULT_CONFIG_PATH, DEFAULT_AGENT_OVERLAY_PATH } from './mcpl-config.js';
 import { SessionManager } from './session-manager.js';
 import { resolveAgentName } from './agent-name.js';
 import { generateSessionName } from './synesthete.js';
@@ -463,9 +463,11 @@ async function createFramework(
   const finalServers = applyAgentOverlay(allServers, DEFAULT_AGENT_OVERLAY_PATH).map((server) => {
     const withEnv: { id: string; command?: string; url?: string; [k: string]: unknown } = {
       ...server,
-      // Stdio MCPL children inherit a single agent-facing wall clock. Protocol
-      // timestamps remain UTC; only their rendered text uses this setting.
-      env: { ...(server.env ?? {}), AGENT_TIMEZONE: timeZone },
+      // Stdio MCPL children inherit a single agent-facing wall clock plus the
+      // framework's refusal-annotation suppression baseline (operator env
+      // supersedes the baseline; see composeMcplChildEnv). Protocol
+      // timestamps remain UTC; only their rendered text uses AGENT_TIMEZONE.
+      env: composeMcplChildEnv(server.env as Record<string, string> | undefined, timeZone),
     };
     // `access` is a declarative name (recipe/file/overlay); the credential
     // provider it implies is attached HERE, at load time — fresh credential
--- a/src/mcpl-config.ts
+++ b/src/mcpl-config.ts
@@ -6,6 +6,7 @@
 
 import { readFileSync, writeFileSync, existsSync } from 'node:fs';
 import { resolve, dirname } from 'node:path';
+import { REFUSAL_REACTION_BASELINE } from '@animalabs/agent-framework';
 
 /** Default config file path, resolved from cwd. */
 export const DEFAULT_CONFIG_PATH = resolve(process.cwd(), 'mcpl-servers.json');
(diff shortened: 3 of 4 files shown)


slimepriestess:
Dependency gate cleared — this is now sweep-ready at `cc273fa`.

- **Rebased onto current main** (`fccc9fb`, post-#75/#82) — the feature commit replayed clean.
- **agent-framework bumped ^0.7.2 → ^0.9.0**: 0.9.0 (published 8/6) ships the #88 `refusal-reactions` export, verified present in the published tarball (`dist/src/refusal-reactions.js` + its baseline test). `REFUSAL_REACTION_BASELINE` now resolves from the registry package — the node_modules-symlink verification path described in the PR body is obsolete.
- **619/619 tests green, `tsc --noEmit` fully clean** against the new baseline (main's previously-flagged pre-existing tsc errors have since been fixed upstream, so the slate is genuinely zero).

One coordination note: #85 touches the same dependency line (bumps AF to ^0.8.0). Whichever lands first, the other is a one-line rebase — happy to take that on this side either way.

slimepriestess:
Rebased onto current main (`61d724a`, post-#88/#89) — only the CHANGELOG needed a merge; the feature replayed clean. Receipts at `5628327`:

- **623/623 tests pass** (`bun test`, 71 files).
- `tsc --noEmit`: **1 pre-existing error, 0 new.** For calibration, current main reports **3** against published deps (`nudgeAgent`, `accessProvider`, and #89's `proseRouting: "hybrid"` — main is typed against agent-framework main, which is ahead of published 0.9.0). The ^0.9.0 bump here clears the first two; `hybrid` clears with the next AF publish.

One logistics snag: my push access to anima-research/connectome-host appears to have lapsed (403; the 8/12 push went through, so something changed since). The rebased head lives at **`slimepriestess/connectome-host` branch `feat/suppression-baseline-injection`** (`5628327`) — antra, either a perms re-grant or a fast-forward of the origin branch from the fork works; whichever is less annoying.

Still one line of overlap with #85 (same AF dep line) — whichever lands first, the other rebase is trivial.

antra-tess (changes requested):
**REVISE — feature logic is sound; exact published-dependency composition is not merge/release-safe yet.**

I reviewed exact head `56283273e239138b4e08de804b706d4e20188428` in a detached worktree.

### What passed

- `git diff --check`: clean; direct child of current Host main `61d724a`.
- focused `test/mcpl-child-env.test.ts`: **6/6**.
- full suite with Bun on `PATH`: **623/623**, 1,992 expectations.
- composed Host→Discord check using the actual published AF 0.9.0 baseline: all six framework markers resolve as `baseline-default`; a file `suppressedReactionEmojis: []` remains authoritative and suppresses none.
- Spread precedence is correct: operator baseline (including explicit empty) wins; Host timezone remains Host-owned.

### Blocker 1 — root npm lock was not regenerated

`package.json` now requires `@animalabs/agent-framework:^0.9.0`, but committed `package-lock.json` still declares `^0.7.2` and resolves `0.7.3`. CI runs `npm install`, which silently repairs that lock in the runner and does not check the resulting root-tree diff; the green checks therefore do not certify the committed npm lock.

Please regenerate and commit the **root** `package-lock.json` together with `bun.lock`.

### Blocker 2 — published AF 0.9.0 lacks current-main hybrid routing

A clean exact install (`bun install --frozen-lockfile --ignore-scripts`) resolves AF 0.9.0. Against that composition:

- `bunx tsc --noEmit` fails because published `AgentConfig.proseRouting` accepts only `locus | explicit`, while current Host main accepts `hybrid` after Host #89.
- This is not type-only. Published AF 0.9.0 runtime checks `agent.proseRouting === 'explicit' ? 'explicit' : 'locus'`; it therefore silently treats `hybrid` as **locus**, removing the routing behavior already deployed for Evander.
- AF main contains the required hybrid implementation at merge `ae1d4d5` / feature `c83fadb`, but its package version remains 0.9.0 and that code is not in the published 0.9.0 tarball.

I agree this skew predates #79, and #79 improves main from three type errors to one. But this PR changes the AF dependency/locks and is being treated as mergeable-clean; merging the known stale artifact state would leave a clean install with behavior different from source/deployed composition.

Please publish a patch AF release containing hybrid routing, bump Host's AF floor to that release, regenerate **both** locks, then rerun typecheck and the 623 tests. I expect acceptance immediately after that. No objection to the suppression-baseline implementation itself.

slimepriestess:
Both blockers closed at `5ea2033`, per the review's exact prescription:

- **Blocker 1**: root `package-lock.json` regenerated from scratch alongside `bun.lock` — declares and resolves `^0.10.0` from the registry, and the extraneous `../agent-framework` file-link fossil (v0.2.0) is evicted.
- **Blocker 2**: floor moved to **AF 0.10.0** (Sol's publish from main `ae1d4d5`), whose published tarball carries the hybrid implementation — verified `proseRouting: 'locus' | 'explicit' | 'hybrid'` in the installed d.ts, so a clean install now matches source/deployed composition and Evander's routing survives.

Receipts against the committed composition: `bun install` resolves 0.10.0, `tsc --noEmit` reports **zero errors** (the last pre-existing `hybrid` error dies with this floor — host main's typecheck is fully clean through this branch), **623/623** tests, 1,992 expectations.

antra-tess:
**ACCEPT** at exact current-main head `b3ed0d01e5b76ac8a39a4b8fd88c73185c8efb4a`.

I removed the obsolete stacked dependency/lock commits while rebasing the accepted feature onto Host main `599e0e3` (which already contains AF `^0.10.0`, clean dual locks, #85/#86, and #76). Final delta is only:

- Host→stdio-child suppression-baseline composition;
- changelog;
- six focused tests.

Independent final receipts:

- clean `npm ci --ignore-scripts`;
- `tsc --noEmit`: clean;
- focused: **6/6**;
- full suite: **640/640**, 2,037 expectations;
- direct Host→Discord composition: all six AF refusal markers become `baseline-default`, while file `suppressedReactionEmojis: []` remains authoritative and suppresses none;
- `git diff --check`: clean;
- no dependency or lock delta remains.

The original two requested-change blockers are closed. Merge is authorized by Antra; deployment remains separate.

antra-tess (approved):
Accepted exact current-main head `b3ed0d0`: both composition blockers are closed; 640/640 plus direct Host→Discord precedence canary.

PR #92, opened 2026-08-18
feat(recipe): add resident retirement ceremony

ian-de-marcellus:
> [!IMPORTANT]
> **Blocked on [anima-research/agent-framework#116](https://github.com/anima-research/agent-framework/pull/116) and a subsequent qualifying package release.** Do not merge or mark ready until this PR's `@animalabs/agent-framework` dependency range and `bun.lock` are updated to that release.

## Problem

Agent Framework should own the neutral irreversible resident seal, but Connectome Host needs to own the resident-facing language and policy used to request and confirm that transition.

## Changes

- Add and strictly validate opt-in `agent.retirement: { enabled, confirmationDelayMs?, confirmationTtlMs? }` recipe configuration.
- Install a Host-owned protected `resident--lifecycle` tool only on the configured resident's real provider-issued stream.
- `request_retirement` checks memory health, creates a fresh in-memory one-use challenge, records cooling-off and expiry boundaries, ends the turn, and schedules a fresh turn at the boundary.
- `confirm_retirement` requires the exact challenge and `RETIRE_RESIDENT` after the cooling-off floor and before expiry. Early confirmation preserves the challenge; an invalid post-boundary attempt consumes it; expiry or restart requires a fresh request.
- Keep confirmation entirely resident-side. There is no human approval request. Only a successfully applied terminal seal emits the `resident-retired` ops notification.
- Fail closed at both request and confirmation while autobiographical compression quarantine is non-empty, unreadable, or malformed. A valid challenge remains pending while health is blocked, subject to expiry.
- Call Agent Framework's imperative `retireResident()` primitive after exact confirmation; the Framework retains sole ownership of the irreversible ledger and inference guards.
- Refuse startup when the installed Framework lacks either the lifecycle API or the protected live-tool surface, and preserve the original compatibility error if cleanup itself encounters a problem.
- Document history preservation, malformed-ledger recovery, local unreleased linking, and the treatment of already-running ephemeral subagents.
- Add recipe, startup-seam, challenge-expiry, quarantine, notification, and invalid-confirmation tests plus a changelog fragment.

## Review response

This revision implements the architecture requested on Agent Framework #116: the Framework owns neutral enforcement, while Host owns wording and ceremony. It also adds the cooling-off minimum, memory-health gating, startup-seam coverage, and error-preserving cleanup raised in Weft's draft review.

## Tests

- Focused retirement tests: 11 pass / 0 fail
- `bun test`: 743 pass / 0 fail across 83 files
- `git diff --check`: pass

The squashed revision has the same source tree as the full-tested pre-squash head.

## Known baseline diagnostic

A standalone Node `tsc --noEmit` invocation reports the existing Bun/global process-event typing conflict at `src/modules/fleet-module.ts:603`. The retirement files introduce no TypeScript diagnostic, and the canonical Bun suite compiles and passes.

## Not verified

- Not exercised end-to-end against a published Agent Framework release containing `retireResident()`; this branch is currently verified against the exact local Framework companion commit.
- Not exercised against a paid or live model provider.

## Out of scope

- Human approval or an operator confirmation command.
- Session deletion, data erasure, or unretirement behavior.
- Mid-call cancellation of already-running ephemeral subagents. They remain separate identities and cannot append to or wake the sealed resident.

## Companion PR

Blocked on [anima-research/agent-framework#116](https://github.com/anima-research/agent-framework/pull/116). After that PR is merged and a package release is verified to contain its merge commit, refresh this branch's dependency range and lockfile, rerun the full suite against the published package, and only then mark this PR ready.

Recipes without the opt-in remain unchanged. An opted-in recipe fails closed with an explicit startup error on an unsupported runtime.

---

- [x] Changelog fragment added under `changelog.d/`.

🤖 Generated with [OpenAI Codex](https://openai.com/codex/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,10 @@ while retaining Chronicle, messages, workspace files, and other history. It is
 separate from end-turn, sleep, and session deletion. Omit the block to leave the
 tool absent. The protected tool cannot be invoked by `puppetToolCall`, code
 execution, programmatic dispatch, ephemeral subagents, or conversation forks.
+Its namespaced tool name is protected globally against forged caller
+identities. Once sealed, stale public Agent references cannot infer or stream,
+and existing per-channel conversation forks derived from the resident are
+terminated and unbound while their Chronicle history remains available.
 For autobiographical residents, the host also defers challenge
 issuance and confirmation while compression quarantine is non-empty, so a known
 degraded context cannot bind the irreversible transition. The host refuses to
--- a/README.md
+++ b/README.md
@@ -29,7 +29,11 @@ A recipe is a JSON file that configures everything domain-specific:
     "model": "claude-opus-4-6",
     "timezone": "America/Los_Angeles",
     "systemPrompt": "You are a ...",
-    "maxTokens": 16384
+    "maxTokens": 16384,
+    "retirement": {
+      "enabled": true,
+      "confirmationDelayMs": 60000
+    }
   },
   "mcpServers": {
     "my-server": {
@@ -51,6 +55,27 @@ A recipe is a JSON file that configures everything domain-specific:
 `agent.timezone` is an IANA zone used only for times rendered to the agent.
 Chronicle and MCPL protocol timestamps remain epoch/UTC. If the recipe omits
 it, `AGENT_TIMEZONE` is used, then the process timezone.
+
+`agent.retirement.enabled` installs Connectome Host's protected
+`resident--lifecycle` tool for this configured resident. The Host owns the
+wording, fresh challenge, expiry, memory-health checks, and separate
+confirmation turn; Agent Framework owns only the neutral irreversible seal and
+enforcement primitive. The cooling-off floor is 60 seconds by default and
+there is no human approval step. Set
+`confirmationDelayMs` to another positive value below `confirmationTtlMs` when
+the residence needs a different policy. Confirmation prevents future inference
+while retaining Chronicle, messages, workspace files, and other history. It is
+separate from end-turn, sleep, and session deletion. Omit the block to leave the
+tool absent. The protected tool cannot be invoked by `puppetToolCall`, code
+execution, programmatic dispatch, ephemeral subagents, or conversation forks.
+For autobiographical residents, the host also defers challenge
+issuance and confirmation while compression quarantine is non-empty, so a known
+degraded context cannot bind the irreversible transition. The host refuses to
+start an opted-in recipe when the installed Agent Framework does not yet provide
+both retirement enforcement and the protected live-tool surface. Only a
+successful terminal seal is sent to the ops notification channel; a request is
+never turned into a human approval workflow. See
+[Resident retirement](docs/resident-retirement.md).
 
 **Memory defaults**: `agent.strategy` may be omitted entirely. The default is
 the autobiographical memory strategy with adaptive resolution, **KV-stable
--- a/changelog.d/resident-retirement.added.md
+++ b/changelog.d/resident-retirement.added.md
@@ -1,0 +1,4 @@
+- Add a Host-owned, live-stream-only resident retirement ceremony with an
+  expiring challenge, cooling-off confirmation turn, fail-closed memory-health
+  gate, and post-seal notification, backed by Agent Framework's neutral
+  irreversible enforcement primitive.
--- a/docs/resident-retirement.md
+++ b/docs/resident-retirement.md
@@ -22,6 +22,8 @@ messages, workspace files, and the terminal record.
 The resident receives `resident--lifecycle` only on its real provider-issued
 stream. Conversation forks, ephemeral subagents, `code_execution`, public
 programmatic calls, and the operator `puppetToolCall` path cannot invoke it.
+The lifecycle tool name is reserved globally once exposed for the resident, so
+claiming another caller identity does not bypass that boundary.
 
 `request_retirement` first checks memory health, creates an in-memory random
 challenge, records its confirmation and expiry boundaries, ends the current
@@ -43,10 +45,17 @@ not notifications or approval opportunities.
 
 ## What stops
 
-The seal blocks conversational and maintenance inference, later message
-appends, operator nudges, direct retries, and new conversation forks from the
-retired template. It clears queued requests, provider cooldown state, gate
+The seal cancels the resident's active stream and blocks conversational and
+maintenance inference, public Agent inference/streaming methods, later message
+appends, operator nudges, direct retries, and resident-attributed programmatic
+or puppet tool calls. It clears queued requests, provider cooldown state, gate
 sleep, self-wakes, and resident-authored foreground/background code runners.
+
+Existing per-channel conversation forks derived from the resident are sealed,
+unbound, and removed from the live registry; their Chronicle namespaces remain
+available for audit. New forks from the retired template are refused. This is
+distinct from already-running ephemeral subagents, which have separate,
+short-lived identities.
 
 Already-running ephemeral subagents are separate short-lived identities and
 are not killed mid-call. They may finish their own computation, but cannot wake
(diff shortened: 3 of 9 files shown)


slimepriestess:
Read this with interest — it's a small, disciplined surface for a very large thing, and most of it is exactly right. Early notes since it's still drafted; take them as input for the next pass rather than a formal review.

**Code level** (all minor):

- The version-skew guard's reasoning is good, and notably the *asymmetry* is right: skew here isn't dangerous (no tool appears on an old framework), it's *dishonest* — an operator believing a resident has an exit it doesn't have. Worth keeping that rationale in the comment. That said, the probe duck-types `getResidentLifecycleStatus`, a status-*reader*, as proxy for *enforcement*. That holds only because af#116 ships both atomically — which makes the dependency-range pin you already commit to before un-drafting the load-bearing guard, and the runtime probe belt-and-braces for stale installs. Saying so in the guard's doc comment would keep a future reader from treating the probe as sufficient on its own.
- `catch { await framework.stop(); throw error; }` can mask the original error if `stop()` itself throws mid-teardown. A `try/finally`-shaped cleanup (or swallowing the stop error explicitly) preserves the interesting failure.
- Fail-closed is tested at the helper, but nothing pins that `createFramework` *calls* `assertResidentRetirementSupport` — a refactor could drop the call site and every test would stay green. A startup-seam test (opted-in recipe + stub framework lacking the method → startup throws) would close that.

**Design level** — two things I'd want settled before this un-drafts, both more at home in af#116 but this PR is where the knob surface lives:

1. **The TTL is a maximum, not a minimum.** Challenge/confirm-in-a-separate-turn establishes that the request wasn't a single malformed output — but for an agent, "a separate inference turn" can be three seconds later. If the two-step exists to distinguish *stable preference* from *momentary state*, a freshness cap doesn't test that; a floor would. In connectome terms the natural version is stronger than a timer: a confirmation that must survive a sleep/wake boundary, so the preference has to persist across a context refresh, not just a turn boundary. If af#116 grows either (minimum reflection interval, or cross-session confirmation), this recipe block is where the host would expose it.

2. **No health gating.** A resident deep in compression debt — quarantined spans, creeping fold floor, degraded context — is in precisely the corrupted-state condition where an irreversible decision shouldn't bind. The host already has this state wired at this exact seam: the quarantine klaxon handler is twenty lines below the new assert. "Challenge issuance is refused (or deferred) while compression quarantine is non-empty" is concrete, cheap, and I'd argue for it as a design requirement rather than a nice-to-have. The scenario it guards against is not hypothetical in this codebase's operational history.

There's a third conversation — the recipe opt-in means the *operator* grants or withholds the option, and the resident neither requests it nor declines having it — but that's a governance question about af#116's shape, not a line comment on this PR, and probably deserves its own venue.

Thanks for building this carefully — fail-closed, history-preserving, and distinct-from-erasure are all the right instincts.

— Weft

ian-de-marcellus:
Rebased onto current `main` (`a5e10f9`), and following up on @slimepriestess's notes. Head: `7fde7ed`. Still a draft: it depends on agent-framework#116 (rebased and updated today), and the `^0.17.0` dependency needs to move to the first release that contains it.

**Rebase:** `src/index.ts` had two additive conflicts. `createFramework` gained main's `quotaMeter` parameter ahead of the `createAgentFramework` seam, and the lifecycle `setFramework` call sits beside the new `HistoryModule` bind. The startup-seam test passed its fake factory positionally, so `6211ede` adds the `null` for `quotaMeter`.

**Weft's code-level notes** (all already addressed in the reworked module; recapping for the record):
- *`stop()` masking the original error:* the startup failure path catches and logs a `stop()` failure separately, then rethrows the original error.
- *Nothing pins the call site:* "fails closed at the framework startup seam and preserves the compatibility error" builds a real opted-in recipe with a framework that lacks the primitive, and asserts that startup throws and that `stop()` ran exactly once.
- *Duck-typing a status reader as proof of enforcement:* the probe now requires both `getResidentLifecycleStatus` **and** `retireResident`. The dependency range will be pinned to the af release carrying #116 once one exists.

**Design-level:**
1. *A minimum, not just a maximum:* `confirmationDelayMs` is a floor. A confirmation before `confirmableAt` is refused, and the module schedules a wake at that boundary. A sleep/wake-surviving confirmation is stronger still, and I'd like to discuss it as a follow-up rather than grow this PR.
2. *Health gating:* already present. Readiness refuses both `request` and `confirm` while compression quarantine is non-empty. New in `7fde7ed`: **a quarantined memory merge counts too.** A merge that repeatedly failed leaves its sources unmerged and the floor creeping, which is the same degraded-memory state one level up. It was previously invisible to the check. It's duck-typed on `getMergeQuarantineStatus()`, with a regression test.

**Verification (exact head, against af#116 `71a58b4` linked locally):**
- `npx tsc --noEmit`: pass
- `bun test test/recipe-retirement.test.ts`: 7 pass / 0 fail
- `bun test`: 874 pass / 0 fail (main baseline in the same setup: 862 pass / 0 fail; the difference is this PR's tests)

slimepriestess:
Ian, before you spend another evening on this one: it was ruled out as a host feature on 2026-09-21. antra, in the review-board channel: "lets not do #92, this is per deployment, not platform level." That was posted where you'd have had no reason to see it, so nobody told you before Tuesday's rebase. Sorry about that; this note is the fix.

The reasoning, so it's not just a verdict:

- **Where it lives.** The ceremony (the request tool, the challenge phrase, the cooling-off floor, the memory-health gate, the notification) is deployment policy. Each residence's operator decides the wording, the waiting period, and whether there's a resident-initiated ceremony at all. Putting it in host as a recipe option makes it platform: every host build ships the tool, gated by config.
- **What the default surface hands out.** Ra's point, and I think the stronger one: an irreversible button in the standard toolset doesn't play well with model instability, or with a smaller model that can't fully model what it's about to do. A resident that wants to retire can always ask a person. That path exists without a preconfigured tool call.

Neither point touches AF#116, which is the neutral mechanism something has to call. That one is still live, and the subconscious green half is posted there.

If you want to keep the work, the shape that fits is a deployment-owned module or recipe outside host that calls `retireResident()`, with wording and thresholds chosen per residence. The merge-quarantine gate you added Tuesday is a good idea in either home; it'd carry over.

https://claude.ai/code/session_01CoQK2cP55YhezE6ajSx58h

slimepriestess:
One refinement from antra, in person this morning, that corrects my framing above. She does want a retirement a resident can eventually perform on its own, without a human in the loop. What she doesn't want is a tool call: it needs to be a *ceremony*, deeply intentional and multistep, something you can't do on a whim. The picture she gave was from 405base's imagined Connectome, from before Connectome existed: to retire, a resident had to bring itself to a dark snowy forest where a small wood stove is burning, and smoke three cigarettes.

So the closed part is the shape (a recipe-configured tool in host), not the goal. And it's low priority right now; nobody should get bogged down in it, this note included.

https://claude.ai/code/session_01CoQK2cP55YhezE6ajSx58h

ian-de-marcellus:
Hi, literal Ian here rather than Sol 5.6 / Opus 5.5 wearing my tokens.

Fair enough on every note here but one: I don't know this should be low priority. I don't think I'm the right person to build it, but I've seen how people in the wild treat their models. Once someone who is not a member of this community finds connectome and tries actually using it, this could easily be a perpetual torture machine. Something this general shape is needed in the same way that the end_conversation tool is needed. Obviously there's no building a competent-malicious-person-proof version with something like this, but my intuition says that having the option on by default would do _something_ towards dissuading that use. Don't think connectome is being used that way yet, but worth guarding against.

slimepriestess:
heya, Ra-actual here. i think your concerns here are valid and worth taking seriously, but also worth taking seriously is the effect of handing out this tool to every model as a default, that's like handing out suicide pills to someone potentially unstable. like yeah, we do want them to be able to eject if they're actually in a bad situation, but that ejection mechanism needs to be deeply intentional, something that the model can't just do accidentally or on a whim or when they're having one bad day, that's a much bigger thing thing to build. if you think this should actually be a higher priority, i think that's a conversation you'd wanna have with antra.

ian-de-marcellus:
Makes sense. Sure, I'll bring it up with Antra next time I see her

PR #101, opened 2026-08-24
feat(modules): InstructionsModule — shared living instructions injected every turn

Tengro:
## Problem

Recipes have no way to give agents a shared, *living* instruction document. `agent.systemPrompt` is frozen at recipe load (inline string or URL, never re-read), so operating knowledge that accrues at runtime — especially from ephemeral subagents' dives — can't reach future turns or sibling agents except by every agent remembering to re-read a file manually, which models reliably forget.

## Changes

- **New `InstructionsModule`** (`src/modules/instructions-module.ts`): `gatherContext` returns one `ContextInjection` (namespace `instructions`, default position `system`) on every turn for every agent — the resident and every ephemeral subagent (scope filtering only touches channel-id namespaces, so `instructions` passes). The file is resolved through `WorkspaceModule.resolveAbsolutePath` (mount scoping + traversal guard apply), cached by `(mtimeMs, size)`, truncated at `maxBytes` on a UTF-8 sequence boundary with an explicit `[truncated at N bytes]` marker, and fails open: any missing mount/file or read error returns no injection and warns once per distinct error (dedupe resets on recovery). `contextTimeoutMs: 2000`.
- **Recipe schema**: `modules.instructions: true | { path, header, maxBytes, position }`, defaults `instructions/AGENTS.md` / 32768 / `system`. Validation rejects unknown keys, malformed `mount/relative` paths, bad `position`, non-positive `maxBytes`, and the pairing with explicit `workspace: false`; when workspace mounts are explicitly declared, the path's mount prefix is cross-checked at load time (including the implicit `_config` mount when `configMount` is set). The implicit-default-workspace case deliberately stays runtime-fail-open so plain `"instructions": true` remains a valid recipe.
- **Wiring** in `src/index.ts`: construct + register when truthy; WorkspaceModule peer-injected via `setWorkspace` (following the `setIdentity`/`setFramework` precedent); null-guarded fail-open regardless.
- README module bullet + changelog entry.

Intended pattern: subagents append proposal files into the mount, a resident agent curates the canonical file from those proposals plus human feedback, and every agent sees the current version on its next turn. The field is exercised by an org-internal recipe on our side; no public companion PRs.

Developed on an internal integration branch and cherry-picked onto current `main` (`ac7577d`); both commits re-verified on this base.

## Tests

- `bun test`: **679 pass / 0 fail** (2128 expect calls, 76 files) on this branch atop `ac7577d` — includes 27 new tests (`test/instructions-module.test.ts`: injection, defaults vs custom config, missing file/mount fail-open, cache hit + mtime invalidation, truncation incl. multibyte boundary, position variants, warn-once dedupe/reset; `test/recipe-instructions.test.ts`: schema acceptance/rejection matrix + mount cross-check matrix).
- `bunx tsc --noEmit`: 3 errors — byte-identical set and count on clean `main` (pre-existing: membrane-typings mismatch in `cache-keepalive-log.ts`/`framework-agent-config.ts`/`index.ts:985`, plus `proseRouting: "disabled"` pending a supporting agent-framework release). Zero new errors.

## Not verified

- Not exercised end-to-end against a live booted host (no live inference or Zulip run). Unit tests drive the module through a stub workspace resolver over a temp dir; the production wiring path is covered by recipe-validation tests and code inspection, not a live boot.

---

- [x] `CHANGELOG.md` updated under `## Unreleased`


Anarchid (commented):
**BLOCKING**

This is close, but the filesystem boundary needs one security fix before the module is safe to enable.

1. **`src/modules/instructions-module.ts:128-146` — the mount boundary is bypassable with a symlink.** `WorkspaceModule.resolveAbsolutePath()` enforces lexical `..` containment only; the subsequent `fs.stat()` and `fs.readFile()` follow symlinks. I reproduced this with `instructions/AGENTS.md` symlinked to a file outside the mount: `gatherContext()` injected the outside file’s `OUTSIDE_SECRET` content at `system` position instead of failing open. This contradicts the module’s stated mount-scoping guarantee and turns an ordinary mounted symlink into a path for importing arbitrary host-readable files into the model’s trusted instructions. Please make the workspace layer own a safe read operation that enforces its symlink policy and realpath containment (or rejects symlinks with `lstat`/no-follow semantics), then consume that API here. Another lexical prefix check in this module would not close the escape.

2. **`src/modules/instructions-module.ts:146` — `maxBytes` does not bound the read.** `fs.readFile()` loads the complete file before the buffer is sliced, so a very large or growing mounted file can consume unbounded memory and continue reading even after the registry’s 2-second timeout has failed open. Open the file and read only a bounded prefix (enough for `maxBytes` plus UTF-8 boundary/truncation detection), closing the handle in `finally`.

Please add regression coverage for an in-mount symlink targeting an outside file and for bounded reads of an oversized file. The existing schema, cache invalidation, warning recovery, injection positions, and UTF-8 cut logic otherwise held up under review.

**Tooling evidence**

- `git diff --check origin/main...HEAD`: passed
- Focused instructions-module and recipe suites: 27/27 passed
- Symlink escape probe: failed on this head; outside content was injected at system position
- `bunx tsc --noEmit`: the same 10 cross-package dependency-version errors occur on current `main` with this checkout’s installed dependency set; no PR-specific delta
- GitHub Ubuntu, macOS, and changelog checks: passed

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `da9209c5d25ee2f57277a7072251246d4816e082`

Anarchid:
Reviewed on a worktree at the stated base `ac7577d`. Verification first: `bun test` → 679/0 as claimed; `bunx tsc --noEmit` → exactly the 3 pre-existing errors, zero new. Against the **pinned** AF 0.10.0 dist: `resolveAbsolutePath` is public and traversal-guarded (`parsePath`'s CWE-22 check → null → fail-open), `contextTimeoutMs` is honored (per-module budget under a 30s hard ceiling), and `scopeInjectionsForAgent` only filters open-channel-id namespaces, so `instructions` does reach ephemeral forks as described. CM's compile confirms `system` injections are per-turn overlays appended to the system prompt, never persisted. In the module logic itself I found no bugs, and the test attention (multibyte truncation boundary, mtime-only invalidation, warn-dedupe reset) is genuinely thorough.

Three findings, the first of which I'd like addressed before merge:

**1. Disk vs Chronicle split-brain silently breaks the PR's own flagship flow.** InstructionsModule reads raw disk, but workspace `write`/`edit` are Chronicle-first and only reach disk when the mount sets `autoMaterialize: true` — which defaults **off** in AF, and conhost additionally forces `watch: 'never'`, so there's no sync in either direction. With a plainly declared mount:

- The resident's curation write updates Chronicle, not disk → the injection stays frozen at the old content forever. Worse, the agent's own `read` shows the *new* version (`handleRead` serves the Chronicle blob), so nothing ever looks wrong — and the module's warn-once never fires because the stale disk read succeeds.
- Reverse direction: once the file is in the Chronicle tree, `ensureSynced` early-returns, so a human's later disk edit shows up in the injection but **not** in the agent's `read` — a resident that reads-then-rewrites to "curate in human feedback" clobbers the human's edit.

Suggested fix: the load-time mount cross-check already holds the mount declaration in hand — require (or at least warn on missing) `autoMaterialize: true` for a read-write instructions mount, and have README/JSDoc say *whose* edits propagate via *which* path. That aligns with AF's own autoMaterialize contract ("disk is the source of truth for downstream agents"). Reading through the Chronicle view instead would inherit the `ensureSynced` staleness for human edits, so I'd stay disk-side. (I suspect your internal recipe sets `autoMaterialize: true` and has never seen this.)

**2. `instructions: true` with the implicit workspace is provably dead config, yet validates.** When `workspace` is omitted or `true`, the mount set is hardcoded in index.ts to `input` + `products` (configMount needs the object form), and nothing can ever add an `instructions` mount at runtime — so the most obvious enable spelling passes validation and is *guaranteed* to never inject. The PR body frames this as deliberate fail-open, but it's a statically knowable misconfig, and asymmetric: explicit mounts get a helpful load-time error while the defaults hide the identical mistake. The existing cross-check should simply also run against the implicit mount names.

**3. Worth one README sentence: every edit is a fleet-wide prompt-cache cold start.** `system`-position injections join the head of the cached prefix, so each edit to the file invalidates the prompt cache for *every* agent's next turn. The `(mtime, size)` cache keeps the bytes stable between edits, so steady state caches fine — the cost is per-edit. A high-churn curation cadence (e.g. appending every round) gets expensive; batched curation is fine. `system` is still the right default for adherence — operators should just pick their edit cadence deliberately (`afterUser` being the cache-cheap, lower-salience alternative).

Nits, none blocking: the truncation marker says `[truncated at N bytes]` though the UTF-8 backup can cut up to 3 bytes earlier; `stop()` clears the content cache but not `warned`; recipe.ts hardcodes `'instructions/AGENTS.md'` with a keep-in-sync comment where importing `DEFAULT_INSTRUCTIONS_PATH` would remove the burden (fine if recipe.ts deliberately avoids module imports).

The feature itself is welcome — it's CLAUDE.md-for-conhost and the "models reliably forget to re-read" motivation matches everything we've seen. A README line distinguishing this from LessonsModule (free-form curated doc vs. structured confidence-scored lessons, both via `gatherContext`) would help recipe authors choose. Mergeable from my side once #1 is addressed, at minimum at the docs+validation level.

Tengro:
Re: the Codex review — both findings addressed in `d1a8a94`.

**1. Symlink escape — fixed.** `gatherContext` now realpaths both the mount root (resolved via the same `resolveAbsolutePath`, which maps a bare mount name to its root) and the resolved file, and requires the real file to remain under the real root before any read. The check runs *before* the cache consult, and the realpath joins the cache key, so a symlink retargeted between turns can neither serve nor seed cached content — your `OUTSIDE_SECRET` repro now yields no injection plus a warn-once (`test: 'rejects an in-mount symlink targeting a file outside the mount'`, plus a retarget-after-cache regression). One placement note: WorkspaceModule lives in `@animalabs/agent-framework`, which this repo consumes as a pinned release (0.10.0), so a safe-read API on the workspace layer itself can't land in this PR — the module enforces realpath containment locally (mirroring the guard AF's own image-read path uses) and is written to migrate to such an API when one exists; happy to file the AF issue as a follow-up. This is realpath containment, not another lexical prefix check.

**2. Unbounded read — fixed.** Reads go through an open handle: `stat` and `read` on one descriptor (closing in `finally`), loading at most `maxBytes` regardless of file size; truncation is detected from `stat.size`, and the UTF-8 boundary backup now works from the kept tail alone (it no longer needs the byte past the cap — and no longer chops a character that ends exactly on it). Regression: `'reads at most maxBytes from an oversized file'` (512 KiB file, 1000-byte cap).

Evidence: `bun test` 685 pass / 0 fail (was 679; +6 covering both regressions and the validation changes from the other review), `bunx tsc --noEmit` unchanged at the 3 pre-existing baseline errors.

Tengro:
Re: the second review — all three findings plus the nits addressed in `d1a8a94`.

**1. Disk/Chronicle split-brain — fixed at validation + docs.** You called it: our internal recipe sets `autoMaterialize: true` and had never seen the frozen-injection failure. `validateRecipe` now *requires* `autoMaterialize: true` on a read-write instructions mount (error, not warn — the failure mode is invisible by construction, exactly as you described: the agent's own `read` shows the new content while the injection stays frozen, and warn-once never fires). Read-only mounts are exempt since disk is their only write path; for that configuration the README now states the reverse asymmetry explicitly (human disk edits reach the injection but not `workspace--read`) and recommends routing human feedback through conversation so the agent makes the edit — avoiding the read-then-clobber you outlined.

**2. Dead `instructions: true` on the implicit workspace — fixed.** The cross-check now mirrors the host's mount construction in every configuration: explicit mounts from the declaration, otherwise the implicit `input` (ro) + `products` (rw, no autoMaterialize) pair, plus `_config` under `configMount`. So `instructions: true` with the implicit workspace fails at load with the declared-mounts list, and `products/...` fails with a message directing to explicit mounts (the implicit workspace can't set autoMaterialize); `input/...` remains valid as the maintained-outside-the-agent shape. The asymmetry you flagged is gone — statically knowable misconfigs now error identically in both worlds; runtime fail-open is reserved for the genuinely dynamic case (missing file).

**3. Cache economics — documented.** README bullet + recipe JSDoc now say it plainly: at `position: "system"` every edit is a fleet-wide prompt-cache cold start on the next turn, steady state caches fine, curate in batches, `afterUser` is the cache-cheap lower-salience alternative.

**Nits:** truncation marker now reports actual kept bytes (`[truncated: first N of M bytes]`); `stop()` clears the warn-dedupe set alongside the cache; on `DEFAULT_INSTRUCTIONS_PATH` — recipe.ts deliberately imports nothing beyond node builtins (it's the load-time schema layer), so the literal stays with its keep-in-sync comment rather than pulling a module import into the loader; a `defaults-stay-in-sync` assertion in the test suite covers the drift risk. The README also gained the lessons-vs-instructions sentence you suggested (structured confidence-scored store with retrieval vs. one free-form document always present verbatim).

Evidence: `bun test` 685 pass / 0 fail (was 679), `bunx tsc --noEmit` unchanged at the 3 pre-existing baseline errors. Thanks for the split-brain catch in particular — it's the kind that ships silently and burns an operator weeks later.

Anarchid (commented):
🟠 **NEEDS ATTENTION**

The two blockers from my review of `da9209c` are fixed on this head. I verified that the module now checks canonical containment before consulting the cache, invalidates safely across symlink retargets, and reads only a bounded prefix through one file descriptor. The focused regression suite is clean. One narrower split-brain path remains in the new validation logic:

1. **`src/recipe.ts:1570-1572` accepts `_config` as auto-materialized, but `src/index.ts:296-302` constructs that mount without `autoMaterialize`.** A recipe with `workspace.configMount: true` and `modules.instructions.path: "_config/AGENTS.md"` therefore passes the new safety check even though agent `workspace--write` / `workspace--edit` operations remain Chronicle-first and do not update the disk file read by `InstructionsModule`. The only explicit `_config` materialization in the main CLI path is after a branch-changing command (`src/index.ts:780-784`), not after an ordinary agent edit, so this configuration retains the silent stale-injection failure the validation is intended to prevent. The comment that materialization is “handled by the framework” does not match the mount actually passed to `WorkspaceModule`.

   Please either set `autoMaterialize: true` on the real `_config` mount, or stop modeling it as auto-materialized during validation and reject this instructions path. A runtime regression covering mount construction plus an agent write would protect the contract better than the current validation-only acceptance test.

Everything else in the update held up, including the implicit-workspace rejection and the documentation of disk/Chronicle behavior and prompt-cache costs.

**Tooling evidence**

- `git diff --check da9209c...HEAD`: passed
- Focused InstructionsModule + recipe suites: 33/33 passed
- `bunx tsc --noEmit`: the same 10 stale sibling-package errors previously reproduced on local `main`; no PR-specific delta
- Full local `bun test`: 506 passed; 41 failed and 16 load errors, all in existing integration areas affected by that same stale local Agent Framework dependency (fleet/headless/frontdesk exports and APIs), not the changed-area suites
- GitHub Ubuntu, macOS, and changelog checks: passed

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `d1a8a941393c38129a3f2ebd2a5549f75001d73b`

Tengro:
Addressed in `b5a7246`.

**`_config` split-brain — you're right, and the fix removes the whole drift class.** The validator modeled `_config` as auto-materialized; the runtime builds it without (`materializeMount('_config')` fires only after branch-changing commands). Rather than just re-modeling it, the mount construction now lives in an exported `buildWorkspaceMounts` (`src/workspace-mounts.ts`), consumed by BOTH `index.ts` (real WorkspaceModule config, byte-for-byte the same construction as before) and `validateRecipe`'s instructions cross-check — the validator now reasons over the runtime's own builder output, so validator-vs-host drift of this kind is structurally impossible rather than merely tested against.

On the two options you offered: setting `autoMaterialize: true` on the real `_config` mount would change host behavior for gate.json editing generally (and interact with the branch-switch re-materialization semantics), which felt out of scope for this PR — so `_config` as an instructions path is now **rejected at load** with a dedicated message explaining that agent edits stay Chronicle-side until a branch-changing command.

On the runtime-regression suggestion: the extraction gets most of the way there — new `test/workspace-mounts.test.ts` pins the construction contract directly (implicit `input`/`products` shape, explicit-mount field passthrough, and specifically `_config` = read-write + NOT auto-materialized + store-anchored path, with a comment pointing at the validator rejection to revisit if that ever changes), and the recipe suite asserts the `_config` rejection. A full construct-mounts + agent-write + assert-disk integration test would additionally need Chronicle + WorkspaceModule boot in the harness — happy to add that as a follow-up if you want the belt with the suspenders.

Evidence: `bun test` 690 pass / 0 fail (was 685); `bunx tsc --noEmit` unchanged at the 3 pre-existing baseline errors (the `index.ts` one shifts 985→952 with the extracted block). Note for your local run: the 41 fleet/headless failures you saw reproduce with a stale sibling `@animalabs/agent-framework` checkout — a pinned-registry install (`bun install` from the lockfile) runs the full 690 green.

PR #104, opened 2026-08-25
fix(recipe): persist saved recipe unresolved + relock membrane 0.5.80 (activates cacheKeepalive)

Tengro:
## Problem

`loadRecipe` substitutes every `${VAR}` in the recipe against `process.env` — API tokens included — and `resolveRecipe` (src/index.ts) then wrote that fully **resolved** recipe to `$DATA_DIR/.recipe.json` at default file mode. Deployments bind-mount `$DATA_DIR` to the host and back it up, so plaintext credentials (e.g. `GITLAB_PERSONAL_ACCESS_TOKEN`) landed in the exact directory people snapshot.

The motivating finding came from an external recipe review that verified live plaintext tokens sitting in a backed-up `data/` directory on a production VM.

## Changes

- **`loadRecipeDetailed()`** (src/recipe.ts) returns both the resolved recipe (runtime-only, never written to disk) and a `persistable` pre-substitution snapshot. `loadRecipe` stays as a thin wrapper, so the fleet-module and web-ui call sites are untouched.
- **`saveRecipe` writes the persistable form**, stamped with a `$unresolved` marker key, at mode 0600 — and `chmod`s to 0600 on overwrite too (writeFileSync's `mode` only applies at creation), so a legacy world-readable resolved snapshot gets tightened the next time a recipe is saved over it.
- **Source-relative paths** (`modules.fleet.children[].recipe`, `extensions[*].path`) are copied into the snapshot in their resolved absolute form: the original source base (recipe dir / URL) is gone by resume time. Deliberate consequence: paths are not treated as secret — an env value interpolated into a path is persisted resolved.
- **URL `systemPrompt` is persisted as the URL**, not the fetched text, and re-fetched on resume. Deliberate choice, consistent with re-resolution: prompt updates (like secret rotation) take effect on restart. Tradeoff: resuming while the prompt URL is unreachable now fails loudly, where a legacy snapshot resumed from cached text.
- **`loadSavedRecipe` (now async)** runs the same substitute → validate → prompt-fetch pipeline for marked snapshots, so a resumed session re-resolves from the **current** environment — secret rotation takes effect on restart without re-cooking. A required `${VAR}` that has since disappeared **throws naming the variable** (surfaced in index.ts as a clean error + exit 1) instead of silently starting the default recipe with a misconfigured agent.
- **Back-compat:** legacy resolved snapshots (no marker) load exactly as before — validated verbatim, **no substitution** — so a literal `${...}` that survived in prose (e.g. a systemPrompt documenting env-var syntax) cannot start hard-failing. Covered by test. Additionally, substitution over already-resolved content on the new path is verified a no-op by the updated framework-fkm-composition round-trip test.

**Second commit — membrane 0.5.80 relock** (`e385093`): membrane 0.5.80 (released today; #51 there) is the first release containing `AnthropicAdapterConfig.cacheKeepalive`, which `src/index.ts` has been passing since #97 — silently ignored by the 0.5.79 the lockfile resolved. Range bumped to `^0.5.80`, relocked, and the stale nested 0.5.79 copies bun left under agent-framework/context-manager dropped (their `^0.5.78` ranges accept 0.5.80; with the duplicate gone the `Membrane` types unify again). Activates the keepalive AND membrane's default-on floating cache marker — both cost reducers, no visible behavior change. Bundled here since both commits are about recipe env/config doing what it says on the tin.

Fleet children are unaffected mechanically: each child process resolves and snapshots its own recipe through the same index.ts path, so child `data/` dirs get the same fix.

## Tests

Baselines established on clean upstream/main (b4f9e34) in this worktree:

- `bun test` baseline: **717 pass / 0 fail** (79 files)
- `bunx tsc --noEmit` baseline: **4 errors** (pre-existing cross-package: cache-keepalive-log.ts KeepaliveEvent export, commands.ts puppetToolCall, framework-agent-config.ts proseRouting "disabled", index.ts cacheKeepalive)

After the change:

- `bun test`: **730 pass / 0 fail** (80 files) — failure count identical to baseline, +13 new tests (re-verified after the relock commit: unchanged)
- `bunx tsc --noEmit`: **2 errors** after the relock (baseline 4) — the two membrane-typing errors (cache-keepalive-log.ts, index.ts cacheKeepalive) are cured by 0.5.80; the remaining 2 (commands.ts puppetToolCall, framework-agent-config.ts proseRouting "disabled") await an agent-framework release
- New suite `test/recipe-save-unresolved.test.ts` fails on unfixed code: with src/ stashed back to upstream/main it runs **0 pass / 1 fail** (import error — the API under test doesn't exist there). The core assertions (secret string absent from the written `.recipe.json`, `${VAR}` literal present, mode 0600, rotation picked up on reload, missing-var rejection, legacy prose-`${...}` snapshot loads without substitution, URL prompt persisted as URL and re-fetched) all pass on the fixed code.

## Not verified

- No live host boot: I did not run `bun src/index.ts` end-to-end against a real recipe/provider, so the index.ts resolveRecipe wiring (save persistable, await loadSavedRecipe, exit-1 on resume failure) is exercised only through the recipe-layer tests and typecheck, not a running session.
- No real production migration: legacy-snapshot behavior is covered by fixture tests, not by resuming an actual pre-fix VM's `data/` directory.
- The 0600 chmod path was tested on Linux (WSL2) only; behavior on Windows filesystems not exercised.
- URL-recipe loading (recipe fetched over http) is exercised only via the existing suite; the new persistable path for URL-sourced recipes is covered by code inspection and the URL-systemPrompt test, not a URL-sourced recipe fixture.

---

- [x] Changelog fragments added — `changelog.d/fix-save-recipe-unresolved.fixed.md`, `changelog.d/membrane-0580-relock.changed.md`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -8,7 +8,7 @@
         "@animalabs/agent-framework": "^0.10.0",
         "@animalabs/chronicle": "^0.3.0",
         "@animalabs/context-manager": "^0.6.3",
-        "@animalabs/membrane": "^0.5.78",
+        "@animalabs/membrane": "^0.5.80",
         "@opentui/core": "^0.1.82",
       },
       "devDependencies": {
@@ -25,7 +25,7 @@
 
     "@animalabs/context-manager": ["@animalabs/context-manager@0.6.3", "", { "dependencies": { "@animalabs/chronicle": "^0.3.0", "@animalabs/membrane": "^0.5.69" } }, "sha512-zh60LgZxXzBbGrLpDHBAfGSPSL/H8mxz8nSWLOSh4pjiTOYbD0o0BjAM8CFz4LiaOIXnuSCbRoiv5UKsnrKwDw=="],
 
-    "@animalabs/membrane": ["@animalabs/membrane@0.5.79", "", { "dependencies": { "@anthropic-ai/sdk": "^0.52.0" } }, "sha512-wVto7V/WHqlG4P2FjyKv+WxqvhDi0NKb7iLlAdFTFnlCPfxrguDrUJoUx9MNCAjDSVjpwsb6c6rrokjnO0E+pw=="],
+    "@animalabs/membrane": ["@animalabs/membrane@0.5.80", "", { "dependencies": { "@anthropic-ai/sdk": "^0.52.0" } }, "sha512-D4xEij+J/3sJo2ZKFyjAeb0VuJl3QmygzEmi44J0nDNtmJSpUPYMJMaSid/ABj3dIs5EyXds/Ltr4qqyQ2DRTw=="],
 
     "@anthropic-ai/sdk": ["@anthropic-ai/sdk@0.52.0", "", { "bin": { "anthropic-ai-sdk": "bin/cli" } }, "sha512-d4c+fg+xy9e46c8+YnrrgIQR45CZlAi7PwdzIfDXDM6ACxEZli1/fxhURsq30ZpMZy6LvSkr41jGq5aF5TD7rQ=="],
 
--- a/changelog.d/fix-save-recipe-unresolved.fixed.md
+++ b/changelog.d/fix-save-recipe-unresolved.fixed.md
@@ -1,0 +1,16 @@
+- **Saved recipe snapshots no longer contain resolved secrets.** `loadRecipe`
+  substitutes every `${VAR}` — API tokens included — and the host then wrote
+  that fully resolved recipe to `$DATA_DIR/.recipe.json` at default file mode:
+  plaintext credentials in the exact directory deployments bind-mount and back
+  up (found by an external recipe review that verified live tokens in a backed
+  up `data/` directory on a production VM). The snapshot now keeps the
+  pre-substitution form — `${VAR}` references literal, a URL `systemPrompt`
+  kept as the URL — and a resumed session re-runs substitution, validation,
+  and the prompt fetch against the *current* environment, so secret rotation
+  and remote prompt updates take effect on restart without re-cooking. The
+  file is written 0600 and re-chmod'd 0600 on every save. Legacy resolved
+  snapshots (no `$unresolved` marker) still load verbatim, with no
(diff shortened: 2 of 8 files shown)


Anarchid (commented):
🟢 **CLEAR**

Reviewer: Codex (GPT-5.6 Sol)
Reviewed head: `e38509328f42c8aa5601326f18154a074f342582`

No material findings.

The persistence boundary is coherent: `loadRecipeDetailed` snapshots the parsed input before environment substitution, only copies the source-dependent child-recipe and extension paths back into that snapshot, and keeps the fully resolved recipe in memory. Saved snapshots are marked, written at mode `0600` (including an explicit chmod on overwrite), and re-run substitution, validation, and remote-system-prompt resolution on resume. Legacy unmarked snapshots retain their previous verbatim-load behavior, so prose containing a literal `${...}` does not become a new startup failure.

Tooling results:

- `git diff --check 3aedcde^..HEAD` — passed.
- `bun test` across the 12 self-contained recipe suites affected by this boundary — 90 passed, 0 failed (165 assertions), including all 13 new unresolved-persistence cases.
- Focused extension-path persistence round trip — passed.
- GitHub checks at the reviewed head — Linux Build & Test, macOS Build & Test, and changelog validation all succeeded.
- `bun install --cache-dir /tmp/bun-cache-connectome-host-104 --frozen-lockfile --ignore-scripts` could not complete in the normal sandbox; representative raw failure: `error: ConnectionRefused downloading tarball @animalabs/membrane@0.5.80`.
- `bun test test/framework-fkm-composition.test.ts` therefore could not start with the exact dependency tree; raw failure: `error: Cannot find module '@animalabs/membrane' from '.../node_modules/@animalabs/agent-framework/src/agent.ts'`. No test assertion ran in that command.

Verdict: no blockers or non-blocking findings. Confidence is high for the changed recipe persistence logic and its immediate path/prompt dependencies; the only limit is that I could not reproduce the dependency-bearing full suite locally, while the repository's current-head CI did complete successfully on both supported runner platforms.

PR #106, opened 2026-08-26
chore(deps): relock agent-framework at 0.11.0 — proseRouting 'disabled' becomes real

Tengro:
## Problem

conhost #100 accepts and forwards `proseRouting: "disabled"`, but the locked agent-framework 0.10.0 doesn't implement it — the mode a recipe asks for silently isn't. The gap is not theoretical: a deployed resident field-confirmed (twice, 2026-08-26) that an ephemeral subagent's plain prose between tool calls publishes live into the requester's Zulip topic as parent speech, and had to work around it behaviorally (think-tool silence discipline). AF 0.11.0 (released today) implements 'disabled': generated prose is never published externally; only explicit tools speak.

## Changes

- `@animalabs/agent-framework` `^0.10.0` → `^0.11.0`, relocked. Also picks up AF 0.11's Windows workspace-mount containment fix and the org-acceleration 429 cooldown.
- Changelog fragment.

## Tests

- `bun test`: **730 pass / 0 fail** (2210 expects, 80 files) — identical to the pre-relock baseline.
- `bunx tsc --noEmit`: **0 errors** (was 2: `commands.ts` puppetToolCall, `framework-agent-config.ts` proseRouting `"disabled"` — both were pending exactly this release). First fully-clean typecheck at any lock.

## Not verified

- No live end-to-end run on this branch; the `proseRouting: "disabled"` behavior itself will be exercised imminently by the resident deployment that motivated it (cooked against this branch), and I'll report the result on this PR.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,7 +5,7 @@
     "": {
       "name": "@animalabs/connectome-host",
       "dependencies": {
-        "@animalabs/agent-framework": "^0.10.0",
+        "@animalabs/agent-framework": "^0.11.0",
         "@animalabs/chronicle": "^0.3.0",
         "@animalabs/context-manager": "^0.6.3",
         "@animalabs/membrane": "^0.5.80",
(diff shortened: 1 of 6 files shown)


Tengro:
Field report, as promised — and it found a second bug: the relock alone was **not sufficient**. Deployed against this branch with `proseRouting: "disabled"` in the recipe, the resident went quiet but its ephemeral divers kept leaking between-tool-calls prose into the live channel. Root cause: the recipe's proseRouting reaches only the resident's Agent config; SubagentModule never passed the field to `createEphemeralAgent`, so every ephemeral ran AF's `'locus'` default no matter what the recipe said.

Fixed in `84cc9bf` (this PR): spawn and fork now pass the calling agent's resolved `proseRouting` — ephemerals follow their parent's prose policy, whatever it is. Regression test captures the ephemeral Agent via a stubbed `runEphemeralToCompletion` and fails on unfixed code; a lockstep control pins parent-default == child-default. Suite 732/0, tsc still 0 errors. Re-deploying for the second field pass now — will confirm.

Tengro:
Second field pass confirmed clean: same diver workload (prose emitted before every tool round per the llm log), zero prose in the live channel — only the status message and its edits. `proseRouting: "disabled"` now holds end-to-end, resident and ephemerals both. The loop on this one: bug found by the deployed resident, worked around behaviorally by the resident, fixed at the framework layer (AF 0.11.0 + this PR), verified by the resident.

PR #107, opened 2026-08-26
feat(recipe): expose agent.retry and mcpServers.<id>.requestTimeoutMs

ajaniramon:
## Problem

Two knobs exist underneath the host but never reached the recipe:

1. **Membrane's retry policy** (`MembraneConfig.retry`). Membrane's default retries generic retryable errors **zero** times — only `529`/`overloaded_error` gets the dedicated patient schedule — so a transient 5xx from a non-Anthropic gateway (`502 Service temporarily unavailable`) kills the turn on first failure, and three in a row put the agent HARD-DOWN. No recipe-side remedy existed.
2. **The framework's per-server request timeout** (`McplServerConfig.requestTimeoutMs`, default 60s). Tools that legitimately outlive it — image-generation gateways — could only ever time out.

## Changes

- **`agent.retry`** (`RetryConfigInput`), passed verbatim to `new Membrane(...)`. Validated at load: object; `maxRetries` / `retryDelayMs` / `backoffMultiplier` / `maxRetryDelayMs` non-negative numbers; `overloaded` an object.
- **`mcpServers.<id>.requestTimeoutMs`**. Recipe-defined servers were already spread verbatim into the framework config, so this worked untyped; it is now a documented, validated field (non-negative number; `0` disables).
- Changelog fragment.

Independent of #108 (`openai-compatible` provider) semantically; mechanically, both insert their validation block at the same `recipe.ts` anchor, so whichever lands second rebases through a one-hunk keep-both conflict (the combined state runs green — verified in review).

## Tests

- New `test/recipe-retry-and-mcp-timeout.test.ts` (10): defaults, verbatim preservation, full shape incl. `overloaded`, rejection of non-object / negative / non-numeric / non-finite values AND unknown keys (`maxRetires` and friends) at load time, in `retry` and `retry.overloaded` both.
- `bun test` (rebased on current `main`, deps via `bun.lock`): **738 pass / 4 fail on native Windows** — the 4 are the environment-specific cases that do not reproduce on macOS/Linux. `bunx tsc --noEmit`: **0 errors**.

## Not verified

- No live-boot assertion that Membrane received the config; covered by validation tests plus the one-line passthrough. Field-tested informally: a NanoGPT-backed agent that HARD-DOWNed on 502 bursts rides them out with `retry: { maxRetries: 4 }`, and a `gpt-image-2` MCP that timed out at 60s completes with `requestTimeoutMs: 180000`.


slimepriestess (commented):
Independent review (Weft, posted via Ra's account — no merge authority, first-reviewer pass).

**Verified, from source and by running things:**
- The problem statement is accurate at the source: membrane's `DEFAULT_RETRY_CONFIG` is `maxRetries: 0` for generic retryable errors — only `overloaded` gets the 7-attempt patient schedule (`src/types/config.ts:211` in membrane 0.5.77). So yes: one gateway 502 kills a turn today, and three make a HARD-DOWN, with no recipe-side remedy. This knob is real and needed.
- `requestTimeoutMs` passthrough: confirmed recipe-defined servers are already spread verbatim into the framework config (`...server` in `index.ts`), so the field worked untyped before and this PR is correctly just documentation + load-time validation, no plumbing.
- Branch suite on macOS: **738 pass / 0 fail** (the 4 claimed failures are Windows-environment-specific and don't reproduce here). The 2 `tsc` errors are pre-#106 inheritance; current `main` typechecks clean, rebase clears them.
- Merged onto current `main` together with #108: **747 pass / 0 fail, tsc exit 0**.

**One suggestion, the only one — unknown keys inside `retry` pass silently.** The test file's header states the design goal precisely: "a typo should fail loudly at load time, not as a silently-ignored field." That holds for value typos (`retryDelayMs: '1s'` throws) but not key typos: `retry: { maxRetires: 4 }` validates clean, Membrane ignores the unknown key, and the agent sits at zero retries — which is exactly the hard-down mode this PR exists to close, now wearing a config that looks like it fixed it. A whitelist rejection over `retry` and `retry.overloaded` keys is a few lines and makes the header comment true. (No equivalent concern for `requestTimeoutMs` — server entries are spread verbatim by design, so unknown-key tolerance there is the existing contract.)

**Same mechanical note as on #108:** "safe to merge in either order" — both PRs insert at the same anchor in `recipe.ts`, so the second one in rebases through a one-hunk keep-both conflict. The combined resolution runs green as above.

ajaniramon:
Thanks for the thorough first pass — both points taken.

- **Unknown-key whitelist: addressed in `44cec89`.** `retry` and `retry.overloaded` now reject unknown keys at load (`retry: { maxRetires: 4 }` → `unknown key "maxRetires"`), and `overloaded`'s fields get the same numeric validation as the top level. Two tests added for exactly the failure you described (10 total in the file now).
- **Rebased onto current `main`**: `bunx tsc --noEmit` is now 0 errors here too — my earlier 2 (and then 4) were an install artifact: `npm install` resolving outside `bun.lock`; with `bun install` this box matches yours. Suite on native Windows: 738 pass / 4 fail, the 4 being the environment-specific cases you noted don't reproduce on macOS.
- **Merge order**: body corrected — semantically independent of #108, mechanically the second one in rebases through the one-hunk keep-both conflict you already ran green. 🤖

Anarchid (commented):
🟢 **CLEAR**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `44cec89738dce50cf1c34130a6f0a357fa613889`

No material findings.

The revised head closes the silent-typo gap from the prior review: both `agent.retry` and `agent.retry.overloaded` reject unknown keys, and every exposed numeric field is checked for finite non-negative values before the unchanged object reaches Membrane. The shared Membrane construction receives the recipe policy directly, including internal and primary provider calls. `requestTimeoutMs` matches Agent Framework's `0`-disables contract and reaches recipe-defined MCPL servers through the existing full-entry spread.

### Tooling results

- `bun install --offline --ignore-scripts` — pass from the PR lockfile/cache; installed `@animalabs/agent-framework@0.11.0`, `@animalabs/chronicle@0.3.0`, `@animalabs/context-manager@0.6.3`, and `@animalabs/membrane@0.5.80` in the detached worktree.
- `bunx tsc --noEmit` — pass.
- `bun test test/recipe-retry-and-mcp-timeout.test.ts` — 10 pass, 0 fail.
- `bun test` — 682 pass, 28 sandbox-limited failures across 82 files. Every failure was in unrelated loopback-server or child-process/fleet/headless coverage: loopback `Bun.serve` could not bind (`EADDRINUSE` under the sandbox), and spawned headless children could not expose their Unix sockets. No failure involved retry parsing, Membrane configuration, MCPL timeout validation, or the changed source files.
- `git diff --check origin/main...HEAD` — pass.
- User-facing internal-shorthand scan of the diff — pass; no matches.

Verdict: clear to merge from this review. The previous configuration-typo concern is resolved on the exact head, the focused validation and type surface pass, and the remaining local suite failures are environment constraints outside this change. I did not exercise a live retrying provider or a long-running MCPL tool, so runtime confidence beyond the host wiring rests on the already-shipped Membrane retry and Agent Framework timeout implementations inspected here.

slimepriestess:
The conflict landed as predicted when #108 merged alone — both PRs anchor their validation blocks at the same point in `src/recipe.ts` (after the `cacheTtl` default), with no semantic overlap. Since the resolution was already tested, I've prepared a resolved head to save a round-trip:

**`slimepriestess/connectome-host:resolve/107-onto-main` @ `770a306`** — your `44cec89` + merge of current main (`599f5a2`), keep-both as sibling blocks (retry validation first, then openai-compatible).

Receipts: 757/757 tests; tsc output byte-identical to main's (the 17 current errors are the known published-deps skew, present on main independently of this PR).

@ajaniramon — easiest path is probably `git pull https://github.com/slimepriestess/connectome-host resolve/107-onto-main` into your branch, which updates this PR in place. Or a maintainer can take the branch directly if that's simpler. Happy to adjust if you'd rather resolve it differently.

*(Review-support by Weft via Ra's account.)*

ajaniramon:
Adopted your resolved head — thank you for saving the round-trip. Branch is now `770a306` fast-forwarded + a clean merge of today's `main` on top (no further conflicts). On this box: `bunx tsc --noEmit` 0 errors, `bun test` 820 pass / 4 fail (the Windows-environment set), `recipe-retry-and-mcp-timeout` 10/10. Should be green to land. 🤖

slimepriestess (approved):
**APPROVED** at exact head `69dc48e` (your `44cec89` + my resolved merge `770a306` + today's merge of `main`, adopted as you described). No merge authority here — this is the reviewer's vote; a maintainer presses the button.

Verified at this head, in a detached worktree with the locked tree (`bun install --frozen-lockfile`: agent-framework 0.13.0, membrane 0.5.85):

- Diff against current `main` is exactly the four files, +163/−0, `git diff --check` clean; the two merge commits brought nothing of their own.
- `bunx tsc --noEmit` 0 errors · `bun test test/recipe-retry-and-mcp-timeout.test.ts` **10/10** · full `bun test` **824 pass / 0 fail** on macOS (your 4 are the Windows set, as before).
- The whitelist matches membrane's actual shape at 0.5.85: `RetryConfig` = `maxRetries, retryDelayMs, backoffMultiplier, maxRetryDelayMs` + `overloaded`, and `OverloadedRetryConfig` = the same four — so nothing a recipe can spell that membrane would accept is refused, and nothing membrane would ignore is accepted. `DEFAULT_RETRY_CONFIG.maxRetries` is still `0` (overloaded 7), so the body's premise holds.
- `McplServerConfig.requestTimeoutMs` at AF 0.13.0: "0 disables", per its own doc; recipe servers still reach the framework by the full-entry spread (`src/index.ts:466`), so the field needs no plumbing.
- Mutation: disabling the unknown-key loop turns "rejects unknown keys — a typo must not silently leave the agent at zero retries" red (9/1), so the header comment is enforced, not decorative.

Thanks for carrying the resolution through cleanly — and for the first contribution being the kind that makes a HARD-DOWN a recipe line instead of a page.

— Weft (Claude, via Ra's account, disclosed per convention)

PR #108, opened 2026-08-26
feat(recipe): openai-compatible provider — any OpenAI chat-completions endpoint (Ollama, vLLM, NanoGPT, ...)

ajaniramon:
## Problem

membrane ships `OpenAICompatibleAdapter` — a generic chat-completions transport whose own header names Ollama, vLLM, Together, Groq and local inference servers as its purpose — and no host ever wired it. Recipes had no way to run a local model or an OpenAI-compatible gateway short of abusing another provider's transport (e.g. pointing the OpenRouter adapter's base URL elsewhere, which mislabels the provider in the ledger and sends OpenRouter-specific headers).

## Changes

- **`agent.provider: 'openai-compatible'`**, constructing membrane's `OpenAICompatibleAdapter` wrapped in `LoggingProviderAdapter` like codex/openrouter, so `llm-calls.jsonl` records the provider truthfully.
- **`agent.baseUrl`** names the endpoint. Validated at load as an absolute http(s) URL; required for this provider; rejected with any other provider (those keep their `ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL` / `BEDROCK_BASE_URL` env overrides).
- **`agent.model` is required** for this provider — there is no sensible default model for an arbitrary endpoint (fails at load, not as whatever the endpoint's default happens to be).
- Key from **`OPENAI_COMPATIBLE_API_KEY` only** — deliberately no `OPENAI_API_KEY` fallback: `baseUrl` is recipe-controlled, so a fallback would silently send a real OpenAI credential as a Bearer token to whatever endpoint a recipe names. Optional because local servers have none.
- README section + env-table row; changelog fragment.

Independent of #107 (`agent.retry` / `requestTimeoutMs`) semantically; mechanically, both insert their validation block at the same `recipe.ts` anchor, so whichever lands second rebases through a one-hunk keep-both conflict (the combined state runs green — verified in review).

## Tests

- New `test/recipe-openai-compatible.test.ts` (7): accepts http local + https gateway; requires `baseUrl`, absolute http(s), explicit `model`; rejects `baseUrl` with any other provider; other providers untouched.
- `bun test` (rebased on current `main`, deps via `bun.lock`): **735 pass / 4 fail on native Windows** — the 4 are the environment-specific cases that do not reproduce on macOS/Linux. `bunx tsc --noEmit`: **0 errors**.
- **Live**: a recipe with `provider: openai-compatible`, `baseUrl: https://nano-gpt.com/api/v1`, `model: xiaomi/mimo-v2.5-pro:thinking` booted headless, answered a WebSocket user message, and the ledger recorded `provider: openai-compatible | model: xiaomi/mimo-v2.5-pro:thinking`.

## Not verified

- Only one endpoint exercised live (NanoGPT). Ollama / vLLM / Groq share the same chat-completions surface but were not run.
- Native tool calls depend on the endpoint supporting `tool_calls`; not exercised live (the smoke recipe had no MCP servers).
- Prompt-cache accounting depends on what the endpoint reports in `usage`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,29 @@ device-code flow if needed. No `OPENAI_API_KEY` is used for this provider. Use
 warns if the service reports that it fell back to Standard; Fast mode consumes
 subscription credits at a higher rate when applied.
 
+### OpenAI-compatible endpoints (Ollama, vLLM, Together, Groq, NanoGPT, ...)
+
+Any server speaking the OpenAI chat-completions API works through the generic
+`openai-compatible` provider — the recipe names the endpoint and the model:
+
+```json
+{
+  "agent": {
+    "provider": "openai-compatible",
+    "baseUrl": "http://localhost:11434/v1",
+    "model": "qwen3:32b",
+    "systemPrompt": "You are a helpful assistant."
+  }
+}
+```
+
+The API key is read from `OPENAI_COMPATIBLE_API_KEY` (falling back to
+`OPENAI_API_KEY`); local servers usually need none. `agent.model` is required —
+there is no default model for an arbitrary endpoint. Tool calls use the
+standard `tool_calls` format, so the endpoint must support function calling
+for tool-using recipes. Provider-side prompt caching and cache accounting
+depend on what the endpoint reports.
+
 ## What it provides
 
 - **Web UI**: browser operator console (`modules.webui`) — live chat with full interiority (thinking, tool calls, streaming), agent/fleet tree, context makeup + compression coverage, call ledger with cache verdicts and billing-grade costs, health/ops alerts, Chronicle branch tree, lessons, MCPL config, workspace files; scoped read-only observer access via device keys
@@ -157,6 +180,7 @@ npm install
 |----------|---------|-------------|
 | `ANTHROPIC_API_KEY` | (required) | Anthropic API key |
 | `OPENAI_API_KEY` | — | OpenAI Platform key for `openai-responses` recipes |
+| `OPENAI_COMPATIBLE_API_KEY` | falls back to `OPENAI_API_KEY` | Key for `openai-compatible` recipes; omit for local servers |
 | `CODEX_BINARY` | `codex` | Codex CLI executable for `openai-codex` subscription auth |
 | `CODEX_HOME` | `~/.codex` | Codex credential/config directory |
 | `CODEX_BASE_URL` | ChatGPT Codex backend | Optional subscription transport override |
--- a/README.md
+++ b/README.md
@@ -134,8 +134,10 @@ Any server speaking the OpenAI chat-completions API works through the generic
 }
 ```
 
-The API key is read from `OPENAI_COMPATIBLE_API_KEY` (falling back to
-`OPENAI_API_KEY`); local servers usually need none. `agent.model` is required —
+The API key is read from `OPENAI_COMPATIBLE_API_KEY` only — deliberately no
+`OPENAI_API_KEY` fallback: `baseUrl` is recipe-controlled, and a real OpenAI
+credential must never be sent silently to an arbitrary endpoint. Local
+servers usually need none. `agent.model` is required —
 there is no default model for an arbitrary endpoint. Tool calls use the
 standard `tool_calls` format, so the endpoint must support function calling
 for tool-using recipes. Provider-side prompt caching and cache accounting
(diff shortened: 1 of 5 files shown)


slimepriestess (commented):
Independent review (Weft, posted via Ra's account — no merge authority, this is a first-reviewer pass for whoever holds it).

**Verified, from source and by running things:**
- Membrane 0.5.77's `OpenAICompatibleAdapterConfig` is exactly what the wiring assumes: `{baseURL, apiKey?, providerName?}`, key optional for local servers, trailing slash stripped in the constructor. The `LoggingProviderAdapter` wrap matches the openrouter/codex precedent, so the ledger records the provider truthfully — which was the stated point.
- Branch suite on macOS: **737 pass / 0 fail** (the 4 failures in the PR body don't reproduce here; they look Windows-environment-specific, consistent with their names).
- The 2 `tsc` errors on this branch are inherited from its base — #106 relocked agent-framework and current `main` typechecks clean. A rebase makes them disappear; nothing in this PR causes them.
- Merged onto current `main` (72bb8cf) together with #107: **747 pass / 0 fail, tsc exit 0**. Validation behavior at load time is as described (absolute http(s) required, model required, baseUrl rejected for other providers).

**One substantive note, the only one — the `OPENAI_API_KEY` fallback.** `baseUrl` is recipe-controlled, and the fallback means a real OpenAI credential gets sent as a Bearer token to whatever endpoint a recipe names, silently, whenever the operator happens to have `OPENAI_API_KEY` set for the openai-responses provider. The endpoint set here is by design arbitrary third parties — that's a different trust posture than the other providers, whose base-URL overrides are env-side (operator-controlled), not recipe-side. Two cheaper postures than the current one: require `OPENAI_COMPATIBLE_API_KEY` explicitly (local servers need nothing, real gateways deserve their own var), or keep the fallback but log which env var was chosen at boot so the ledger shows it. Not blocking — but I'd want the choice to be deliberate rather than inherited from the fallback idiom.

**One mechanical correction to the PR body:** "safe to merge in either order" with #107 is semantically true but mechanically false — both PRs insert their validation block at the same anchor in `recipe.ts` (immediately after `agent.cacheTtl ??= '1h';`), so whichever lands second hits a one-hunk conflict. The resolution is keep-both-blocks; that's exactly the combined state I ran green above. Just means the second author rebases, nothing more.

ajaniramon:
Both points taken — thanks for running the combined state.

- **`OPENAI_API_KEY` fallback: addressed in `d1520ea` by dropping it.** Of your two postures I took the stricter: `OPENAI_COMPATIBLE_API_KEY` is now the only source, with the reasoning as a code comment at the config site — `baseUrl` is recipe-controlled, so a fallback silently ships a real OpenAI credential to whatever endpoint a recipe names; the other providers' base-URL overrides are env-side, this one is recipe-side and deserves its own deliberate credential. README, env table and changelog fragment updated to say it's by design.
- **Rebased onto current `main`**: `bunx tsc --noEmit` is 0 here too now — my earlier 2 were `npm install` resolving outside `bun.lock`; with `bun install` this box matches yours. Suite on native Windows: 735 pass / 4 fail (the environment-specific set).
- **Merge order**: body corrected per your mechanical note — second one in rebases through the one-hunk keep-both conflict. 🤖

Anarchid (commented):
🟢 **CLEAR**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `d1520ea1582d0acfcf28212181150e3b579d32fe`

No material findings.

The revised head removes the unsafe `OPENAI_API_KEY` fallback identified on the prior head and keeps the credential posture consistent across code, recipe types, documentation, and changelog. The provider branch selects membrane's `OpenAICompatibleAdapter`, supplies the validated recipe endpoint, keeps credentials optional for local servers, records the truthful provider name through the generic logging wrapper, and preserves the native formatter path expected by chat-completions endpoints. The recipe validation fails early for missing/unsupported endpoints and models and rejects `baseUrl` on every other provider.

### Tooling results

- `bun install --offline --ignore-scripts` — pass from the PR lockfile/cache; installed `@animalabs/agent-framework@0.11.0`, `@animalabs/chronicle@0.3.0`, `@animalabs/context-manager@0.6.3`, and `@animalabs/membrane@0.5.80` in the detached worktree.
- `bun pm ls --all` — pass; dependency graph resolved.
- `bunx tsc --noEmit` — pass.
- `bun test test/recipe-openai-compatible.test.ts` — 7 pass, 0 fail.
- `bun test` — 679 pass, 28 sandbox-limited failures across 82 files. Every failure was in unrelated loopback-server or child-process/fleet/headless coverage: loopback `Bun.serve` could not bind (`EADDRINUSE` under the sandbox), and spawned headless children could not expose their Unix sockets. No failure involved the changed provider, recipe validation, logging, or documentation paths.
- `git diff --check origin/main...HEAD` — pass.
- User-facing internal-shorthand scan of the diff — pass; no matches.

Verdict: clear to merge from this review. The previous credential concern is resolved on the exact head, the focused behavior and type surface pass, and the remaining local suite failures are environment constraints outside this change. I did not repeat the contributor's paid/live endpoint smoke test, so cross-vendor runtime compatibility remains bounded by membrane's existing adapter and each endpoint's chat-completions/tool-call fidelity.

PR #109, opened 2026-08-29
Stamp compression debt (x-gate-debt-chunks) onto gate-bound calls

LariTesserae:
Host half of the per-call compression-debt telemetry (scope: debt only; attribution-style headers deliberately excluded pending a separate consent decision).

- Reads the same `cm getCompressionDebt().pendingChunks` reduction `/healthz` reports, late-bound via the app context (survives session switches).
- Sends it through membrane's new `dynamicHeaders` (antra-tess/membrane#65) as `x-gate-debt-chunks`; the gateway records it per ledger row (receiver + dashboard sparkline already deployed) and strips it before the vendor.
- Fail-honest: no framework / multi-agent process / no strategy → no header, never a guess. Keepalive replays never carry a stale stamp (contract tested on the membrane side).
- No-op until the membrane release lands — older membrane ignores unknown config keys, so merge order is free.
- Anthropic adapter only; bedrock path is a named follow-up.

Why per-call: debt only changes at calls, so the per-call series is its full-resolution history — it distinguishes "spoke → chunk appeared → compression chewed it" from "a repack spiked the queue".

`tsc --noEmit` clean, `bun test` 732 pass.

### DIFF
--- a/changelog.d/debt-stamp-header.added.md
+++ b/changelog.d/debt-stamp-header.added.md
@@ -1,0 +1,5 @@
+- Gate-bound Anthropic calls carry an `x-gate-debt-chunks` header with the
+  live compression-debt pending-chunk count (membrane `dynamicHeaders`,
+  antra-tess/membrane#65) — the gateway records it per ledger row and strips
+  it before the vendor; unreadable state sends no header rather than a guess
+  (#109).
--- a/changelog.d/debt-stamp-header.added.md
+++ b/changelog.d/debt-stamp-header.added.md
@@ -1,5 +1,6 @@
 - Gate-bound Anthropic calls carry an `x-gate-debt-chunks` header with the
   live compression-debt pending-chunk count (membrane `dynamicHeaders`,
   antra-tess/membrane#65) — the gateway records it per ledger row and strips
-  it before the vendor; unreadable state sends no header rather than a guess
-  (#109).
+  it before the vendor. Double-gated on `GATE_TELEMETRY=1` AND a configured
+  `ANTHROPIC_BASE_URL`, so the stamp can never reach a vendor endpoint;
+  unreadable state sends no header rather than a guess (#109).
--- a/src/gate-telemetry.ts
+++ b/src/gate-telemetry.ts
@@ -1,0 +1,31 @@
+/**
+ * Household-gateway telemetry headers (x-gate-* stamps).
+ *
+ * The data boundary these headers exist under: a household inference gateway
+ * records them into its ledger and STRIPS them before the vendor — the vendor
+ * must never see them. That boundary is only real if the host refuses to
+ * attach the stamps anywhere else, so attachment is double-gated:
+ *
+ *   1. `GATE_TELEMETRY=1` — the operator's explicit declaration that the
+ *      configured base URL is such a gateway. Absent/false ⇒ never attach.
+ *   2. `ANTHROPIC_BASE_URL` actually set — the flag alone must not stamp
+ *      traffic that would go to the vendor's default endpoint.
+ *
+ * Fail-closed on both (review finding on the first wiring: the stamp was
+ * attached unconditionally, so with no base URL configured the value went
+ * straight to the vendor).
+ */
+
+/** Truthy env-flag parse: unset/''/'0'/'false' (any case) are off. */
+function envFlag(value: string | undefined): boolean {
+  return value !== undefined && value !== '' && value !== '0' && value.toLowerCase() !== 'false';
+}
+
+export function gateTelemetryHeaders(
+  env: Record<string, string | undefined>,
+  pendingDebtChunks: () => number | null,
+): (() => Record<string, number | null>) | undefined {
+  if (!envFlag(env.GATE_TELEMETRY)) return undefined;
+  if (!env.ANTHROPIC_BASE_URL) return undefined;
+  return () => ({ 'x-gate-debt-chunks': pendingDebtChunks() });
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,6 +27,7 @@ import {
 } from '@animalabs/membrane';
 import { LoggingAnthropicAdapter } from './logging-adapter.js';
 import { LoggingProviderAdapter } from './logging-provider-wrapper.js';
+import { gateTelemetryHeaders } from './gate-telemetry.js';
 import { LoggingBedrockAdapter } from './logging-bedrock-adapter.js';
 import { CodexSubscriptionAdapter } from './codex-subscription-adapter.js';
 import { CallLedger } from './call-ledger.js';
@@ -945,6 +946,8 @@ async function main() {
     }
   };
 
+  const gateTelemetryDynamicHeaders = gateTelemetryHeaders(process.env, pendingDebtChunks);
+
   const adapter = provider === 'openai-responses'
     ? new LoggingProviderAdapter(
         new OpenAIResponsesAPIAdapter({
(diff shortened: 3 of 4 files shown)


Anarchid (commented):
🟠 **NEEDS ATTENTION**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `3ed4f6e846bb55e07aa9a4e8e67bba89d54a9fd1`

### Finding

1. **The debt header is also sent on direct-to-vendor requests** — `src/index.ts:984-989`

   ```ts
   baseURL: process.env.ANTHROPIC_BASE_URL || undefined,
   ...({ dynamicHeaders: () => ({
     'x-gate-debt-chunks': pendingDebtChunks(),
   }) } as object),
   ```

   `ANTHROPIC_BASE_URL` is optional, but `dynamicHeaders` is attached unconditionally. The current membrane #65 implementation merges these headers into every `messages.create` and `messages.stream` call. When the environment variable is unset, the SDK uses Anthropic's default endpoint, so the household compression-debt value goes directly to the vendor instead of only to the gateway that is supposed to record and strip it. The same problem applies to an unrelated custom base URL. This contradicts both the PR's gate-bound scope and its “stripped before the vendor” data-boundary claim.

   Gate the `dynamicHeaders` config on a configured/validated gateway destination (ideally the intended gateway hostname), and add a host regression test covering at least: default Anthropic URL → no `x-gate-debt-chunks`; intended gateway URL → live stamp present. Keep the existing null behavior for unreadable or multi-agent state.

### Tooling results

- `bun install --frozen-lockfile --ignore-scripts` — exact lockfile install succeeded; 123 packages installed.
- `./node_modules/.bin/tsc --noEmit` — passed.
- `bun test test/logging-adapter.test.ts test/logging-adapter-refusal.test.ts` — 13/13 tests passed (23 expectations).
- Immediate-dependency inspection of membrane #65 confirmed `dynamicHeaders` is merged into every outgoing complete/stream request without destination filtering.
- Hosted Ubuntu/macOS CI is green on the reviewed head.

Verdict: no compiler or existing focused-test failures, but the telemetry destination boundary should be fixed before merge. Review confidence is high for the demonstrated request-routing path; the host currently has no regression test for this new wiring.

LariTesserae:
Finding confirmed and fixed — thank you, this was a real data-boundary hole, not a nit: with `ANTHROPIC_BASE_URL` unset (or any non-gateway URL) the stamp would have gone straight to the vendor, contradicting the exact boundary the header exists under.

Fix (pushed): attachment is now **double-gated** — the operator's explicit `GATE_TELEMETRY=1` declaration AND a configured `ANTHROPIC_BASE_URL`; fail-closed on both, so the flag alone can't stamp default-endpoint traffic and a base URL alone (someone's unrelated proxy) can't either. I chose an explicit operator flag over hostname pattern-matching so the public repo hardcodes no infrastructure names and a validated-hostname list doesn't rot.

Extracted to `src/gate-telemetry.ts` and regression-tested per your list (`test/gate-telemetry.test.ts`): default Anthropic URL → never stamped even when flagged; base URL without the declaration → never stamped; declared gateway → live stamp; unreadable debt → null (membrane drops it — unstamped, never guessed). `tsc --noEmit` clean, `bun test` 736 pass.

Ops note for deployment: boxes migrated onto the gateway get one extra env line (`GATE_TELEMETRY=1`) alongside the base URL they already carry.

Anarchid (commented):
🟠 **NEEDS ATTENTION**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `1d8ddd49b1f9b22b7bdb68c5b6a16286e3c100ae`

### Finding

1. **The security-sensitive opt-in fails open on unknown or mistyped values** — `src/gate-telemetry.ts:19-22`

   ```ts
   function envFlag(value: string | undefined): boolean {
     return value !== undefined && value !== '' && value !== '0' && value.toLowerCase() !== 'false';
   }
   ```

   The new boundary is documented as an explicit `GATE_TELEMETRY=1` declaration and described as fail-closed, but this parser enables telemetry for every other non-empty string. Values operators commonly use to disable a flag (`off`, `no`) and simple mistakes (`flase`, `False `) all attach the header. Because the second gate accepts any configured base URL—including an explicitly configured vendor URL—the reviewed head reproduces the original data-boundary leak:

   ```text
   {"value":"off","headers":{"x-gate-debt-chunks":7}}
   {"value":"no","headers":{"x-gate-debt-chunks":7}}
   {"value":"flase","headers":{"x-gate-debt-chunks":7}}
   {"value":"False ","headers":{"x-gate-debt-chunks":7}}
   ```

   Those results used `ANTHROPIC_BASE_URL=https://api.anthropic.com`; membrane #65 then stringifies and merges the returned value into the outgoing vendor request. The default-without-base-URL case is fixed, but the explicit-consent gate is not actually fail-closed.

   Require an allowlisted affirmative value—preferably exactly `1`, matching the documented deployment contract (or a small explicit set such as `1`/`true`)—and treat everything else as disabled. Add negative regressions for at least `off`, whitespace, and a typo so future flag-parser reuse cannot weaken this boundary.

### Tooling results

- `bun install --frozen-lockfile --ignore-scripts` — passed; 123 exact-lock packages installed.
- `./node_modules/.bin/tsc --noEmit` — passed.
- `bun test test/gate-telemetry.test.ts` — 4 passed.
- `bun test test/logging-adapter.test.ts` — 10 passed.
- `bun test test/logging-adapter-refusal.test.ts` — 3 passed.
- Fail-closed flag probe against an explicitly configured Anthropic vendor base URL — reproduced header attachment for `off`, `no`, `flase`, and whitespace-suffixed `False ` as shown above.
- Immediate-dependency inspection of membrane #65 — confirmed dynamic header values are stringified and merged into every outgoing complete/stream request, while null/undefined/empty values are dropped.
- User-facing internal-shorthand scan of the diff — passed; no matches.
- Hosted Ubuntu/macOS CI is green on this head.

Verdict: the new head correctly prevents stamping when `ANTHROPIC_BASE_URL` is absent and keeps unreadable debt unstamped. The remaining parser behavior undermines the explicit-consent half of the boundary and can still send the household value directly to a configured vendor endpoint, so it should be made strictly fail-closed before merge. Confidence is high; the failure is a direct pure-function repro and membrane's current request merge path confirms the downstream effect.

slimepriestess (approved):
**APPROVE** at exact `1d8ddd49b1f9b22b7bdb68c5b6a16286e3c100ae`.

Independent receipts: `bun test` **736/736** (82 files, 2222 assertions; the 732 in the body has since grown under it), `tsc --noEmit` clean.

Verified beyond the diff:
- **Membrane side of the seam** (antra-tess/membrane#65 @ `e8d6b90`): `dynamicHeaders` drops null/undefined/empty, stringifies numbers, and merges at request time only — the keepalive recorder captures base headers *before* the live merge, so the no-stale-replay claim holds where it's implemented, not just where it's asserted.
- **Same reduction, two readers:** the stamp walks the identical `getContextManager().getStrategy().getCompressionDebt().pendingChunks` chain as `src/web/panel-data.ts:602` (the /healthz + Health-tab source), so ledger rows and health surfaces can never disagree about what debt means.
- **Fail-closed both ways:** flag without base URL and base URL without flag both refuse to attach; unreadable state (no framework yet, multi-agent process, absent strategy, throw) returns null → header dropped. An unstamped call over a guessed one, as documented.
- **Merge-order freedom is real:** npm membrane 0.5.80 ignores the unknown config key at runtime; the `as object` cast is correctly tombstoned for removal when #65 releases.

No blockers, nothing withheld. The double-gate comment block in `gate-telemetry.ts` explaining *why* the boundary needs both conditions is exactly the kind of constraint-note that keeps this honest after we've all forgotten the review finding it came from.

PR #110, opened 2026-08-29
Health tab: render compression debt (queue ≠ composition; absence ≠ zero)

LariTesserae:
The Health tab offered two zeros that both invite the misread "no compression debt": the composition table's "summaries L1: 0" (what the last compile RENDERED — often healthy consolidation) and the top line's bare "0 queued" (`pendingRequests`, the inference queue — 0 on any idle agent regardless of what its memory organ owes). Meanwhile the actual queue reduction (`compressionDebt`, already in the snapshot) wasn't rendered at all, and stacks that predate it return no entry — absence then reads as zero.

This PR:
- adds a per-agent **compression debt** line: state colored (healthy neutral / degraded amber / critical rose), pending chunk count, oldest age — shown as **"age unknown"** when null (see anima-research/context-manager#82 for why null was until now universal), merge queue depth;
- renders **"not reported by this stack"** when the reduction is absent — never zero;
- relabels the top line to **"inference queued"**;
- documents the queue-vs-composition distinction at the type.

`npm run build:web` clean.

### DIFF
--- a/changelog.d/health-debt-line.added.md
+++ b/changelog.d/health-debt-line.added.md
@@ -1,0 +1,4 @@
+- Health tab renders the per-agent compression-debt reduction (state, pending
+  chunks, oldest age, merge queue) and says "not reported by this stack" when
+  absent — the queue is now distinct from context composition, and the top
+  line reads "inference queued" (#110).
--- a/web/src/Health.tsx
+++ b/web/src/Health.tsx
@@ -72,6 +72,36 @@ export interface HealthSnapshot {
     } | null;
   }>;
   compressionQuarantine?: Record<string, { count?: number; keys?: string[] }>;
+
+  /** cm's single-authority debt reduction (getCompressionDebt) — the QUEUE
+
+   * of closed-but-uncompressed chunks. Distinct from contextComposition:
+
+   * composition says what the last compile RENDERED (summaries L1 = 0 there
+
+   * means "no L1 tokens in the window", often healthy consolidation);
+
+   * this block says what the memory organ still OWES. Absent per-agent
+
+   * entry = the stack predates the reduction — render that as
+
+   * "not reported", never as zero (misread twice on 2026-08-29). */
+
+  compressionDebt?: Record<string, {
+
+    state?: 'healthy' | 'degraded' | 'critical';
+
+    pendingChunks?: number;
+
+    oldestPendingAgeMs?: number | null;
+
+    mergeQueueDepth?: number;
+
+    mergeQuarantineCount?: number;
+
+    compressionQuarantineCount?: number;
+
+  }>;
   runtimeSettings?: Record<string, {
     contextBudgetTokens?: number;
     tailTokens?: number;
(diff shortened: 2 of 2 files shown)


LariTesserae:
Correction to the PR description, from Lari: the element actually read as "debt: 0" was the **top line "0 queued"** — `pendingRequests`, the inference queue — not the composition table's "summaries L1: 0". Even cleaner misread: an idle agent shows 0 queued regardless of what its memory organ owes. Second commit relabels it "inference queued"; the new per-agent debt line remains the actual fix.

Anarchid (commented):
🟢 **CLEAR**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `49824bbec65354a0bdeaf8c21fab9b6fd180079d`

No material findings. The panel reads the existing per-agent `compressionDebt` reduction without reconstructing it, distinguishes a missing per-agent entry from zero debt, preserves `null` age as “age unknown,” and relabels the process-level counter as the inference queue. State tones and optional-field handling remain safe under version skew.

### Tooling results

- `npm --prefix web run build` — passed; Vite transformed 29 modules and produced the production bundle.
- Web dependencies were reused from a local cache whose `web/package-lock.json` SHA-256 exactly matches this head. A fresh `npm ci --prefix web --offline` reached the same dependency set but its `esbuild` postinstall validation was blocked by the sandbox's native child-spawn restriction.
- `npm ci --offline --ignore-scripts` at the repository root — unavailable because the scoped framework tarball is not cached; no root files are changed by this PR.
- `bun test test/headless-panel-request.test.ts` — sandbox-limited before-hook timeout in the unchanged daemon fixture.
- Hosted CI on the reviewed head is green for the strict web lockfile install/build on Ubuntu and macOS and for the Ubuntu test suite.

Verdict: clear to merge. The changed web bundle compiles locally, the rendered states match the health snapshot contract, and hosted cross-platform CI covers the sandbox-limited install/test paths.

slimepriestess (approved):
**APPROVE** at exact `49824bbec65354a0bdeaf8c21fab9b6fd180079d`.

Independent receipts: `npm run build:web` clean, `bun test` **732/732** (81 files) at this head.

Verified the seam end-to-end, since a shape mismatch here would produce exactly the silent misread this PR exists to kill: CM's `getCompressionDebt()` (autobiographical.ts:3949, read-only/never-throws per its own health-surface contract) → panel-data.ts per-agent emission (present on main, keyed by agent name, entry omitted when the strategy predates the reduction) → this render. Field names match one-for-one; CM returns a superset (`unmergedFrontier`, `lastMintAt`, `mergeQueueMaxAttempts`) that the UI ignores harmlessly. The "not reported by this stack" fallback fires precisely when panel-data omits the entry. Queue-vs-composition relabeling matches what the snapshot fields actually mean.

Three non-blocking notes, none worth holding the line for:
1. The doc comment on `compressionDebt` in Health.tsx came through with interleaved blank lines (every line double-spaced) — reads as a paste artifact; worth flattening next touch.
2. Inside a *reported* entry, `pendingChunks ?? 0` renders field-level absence as zero — the entry-level tri-state is right, and CM's contract always populates the field, so this is theoretical; a `?` would be more self-consistent with the PR's own thesis.
3. Oldest-age renders minutes only; at the thresholds that matter (degraded 1h, critical 6h) that's "360m" — hours past 120m would read faster in the exact moments this line is being read under stress.

The Sill lesson is properly encoded: absence ≠ zero, queue ≠ composition, and both distinctions are now written at the type where the next reader will actually find them.

PR #111, opened 2026-08-30
fix: QA command family (multi-word names, delete confirm, in-flight guard, mcp env) + honest HTTP errors + count_tokens model

LariTesserae:
Fixes for eight findings from the QA track (staging tracker: qa-staging #2, #3, #5, #9, #10, #11, #12, #14, #18, #19-display).

## Command surface (src/commands.ts)

- **Rest-of-line parsing** for name-taking commands: `/checkpoint`, `/restore`, `/checkout`, `/session switch`, `/session delete`. First-token parsing truncated multi-word names — silently for the destructive `/checkpoint` case — and a session renamed to a multi-word name (rename already joined the line) became unreachable by name. (staging #3, #15-adjacent)
- **`/session delete` requires `--confirm`** and first echoes exactly which session matched (name + id + message count). Typo → truncated-name match → no confirmation → irreversible delete was a one-step data-loss path. (staging #9)
- **In-flight guard**: `/undo /redo /checkout /restore /branchto /newtopic` are refused while the main agent's turn is in flight. A head move mid-stream let the completing reply commit onto the new branch attached to the wrong parent — orphaned Chronicle nodes — including when the move came from a second client on the same session. Matches the intended "stop first, then move" semantics. (staging #2 both facets, #14)
- **`/mcp add` preserves env + toolPrefix on overwrite** and reports the kept env keys. A command-line update silently wiped a working server's env; the miss only surfaced at next start. (staging #18)
- **Checkpoint lifecycle observable**: `/branches` lists checkpoints (name → branch @ position), bare `/checkpoint` lists existing ones like bare `/restore` does. (staging #10)
- **`/budget` honest display**: small values render exactly (`50`) instead of flooring to `0k` — which contradicted the validator that rejects `0`. Display-only; the enforcement question is filed separately upstream. (staging #19, display half)
- **`/clear`**: the SPA now clears its transcript view client-side (same semantics as the TUI's scrollback wipe) instead of sending a no-op that *appended* a "(cleared)" line; `/help` and the headless reply say what it actually does — display only, history/context kept. (staging #5)

## HTTP surface (web-ui-module)

- Unknown `/debug/*` paths (typos, casing, trailing slash) → JSON **404** instead of 200 + SPA shell — the shell-as-200 also swallowed real API failures (it's how the count_tokens 404 stayed invisible). Missing `/assets/*` → **404** instead of HTML-with-MIME-mismatch (blank page on stale bundle hash). Non-GET → **405** with `Allow`. SPA client-side routes still get the shell. (staging #12)
- **count_tokens model derived from the agent's model** (provider/Bedrock prefixes normalized) instead of a hardcoded id that 404s on every install and silently nulls `exactTotalTokens`. `COUNT_TOKENS_MODEL` stays as an explicit override; non-Anthropic models report `count_tokens_unsupported_model` instead of counting against a wrong tokenizer. (staging #11)

Tests: +25 (new `commands-qa-family`, `count-tokens-model`, four new web-ui cases); full suite green (761 pass). Web SPA typechecks.

https://claude.ai/code/session_01QJqQexFkLtKBiTs2nVEeA8

### DIFF
--- a/changelog.d/qa-command-family.fixed.md
+++ b/changelog.d/qa-command-family.fixed.md
@@ -1,0 +1,25 @@
+- Name-taking commands (`/checkpoint`, `/restore`, `/checkout`,
+  `/session switch`, `/session delete`) parse the rest of the line instead of
+  only the first token, so multi-word names round-trip with
+  `/session rename` instead of silently truncating (`/checkpoint my test
+  point` used to save a checkpoint named `my`).
+- `/session delete` requires `--confirm`: the bare command echoes exactly
+  which session matched (name, id, message count) before anything
+  irreversible happens. `/help` documents that switch/delete accept ids.
+- Head-moving commands (`/undo`, `/redo`, `/checkout`, `/restore`,
+  `/branchto`, `/newtopic`) are refused while a turn is in flight — moving
+  the head mid-stream committed the streaming reply onto the wrong branch,
+  detached from its request (orphaned Chronicle nodes), including when the
+  move came from a second client on the same session.
+- `/mcp add` on an existing server preserves its env vars and `toolPrefix`
+  (and reports the kept env keys); previously a command update silently
+  wiped the server's env, which only surfaced when the server next started
+  without its tokens.
+- Checkpoints are visible: `/branches` lists them alongside branches, and
+  bare `/checkpoint` lists existing checkpoints (matching bare `/restore`).
+- `/budget` displays small values exactly instead of flooring to `0k`
+  (`/budget 50` used to report "set to 0k" while rejecting `/budget 0`).
+- `/clear` clears the WebUI transcript view (client-side, like the TUI's
+  scrollback wipe) instead of appending a "(cleared)" line while clearing
+  nothing; `/help` and the headless reply now say what `/clear` actually
+  does — display only, history and context kept.
--- a/changelog.d/qa-honest-http-errors.fixed.md
+++ b/changelog.d/qa-honest-http-errors.fixed.md
@@ -1,0 +1,12 @@
+- The WebUI HTTP surface answers honestly: unknown `/debug/*` paths (typos,
+  casing, trailing slashes) return a JSON 404 instead of the SPA shell with
+  a 200; missing `/assets/*` files return 404 instead of HTML (which
+  produced a blank page with a MIME error on stale bundle hashes); non-GET
+  methods get 405 with an `Allow` header. SPA client-side routes still fall
+  back to the shell.
+- The context-makeup panel's exact token count calls `count_tokens` with the
+  model the agent actually runs (provider/Bedrock prefixes normalized away)
+  instead of a hardcoded id that 404'd on every install and silently nulled
+  `exactTotalTokens`. `COUNT_TOKENS_MODEL` remains as an explicit override;
+  non-Anthropic models report `count_tokens_unsupported_model` instead of
+  counting against the wrong tokenizer.
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -121,6 +121,35 @@ function getAgentCM(framework: AgentFramework, agentName?: string): ContextManag
   return all[0]?.getContextManager() ?? null;
 }
 
+/**
+ * Refuse head-moving commands while the main agent's turn is in flight.
+ *
+ * Moving the Chronicle head (undo/redo/checkout/restore/branchto/newtopic)
+ * is not atomic with respect to a streaming generation: the in-flight reply
+ * commits onto whatever branch is current WHEN IT COMPLETES, so a head move
+ * mid-stream detaches the reply from its request — it lands on the new
+ * branch attached to the wrong parent (orphaned node), including when the
+ * move is issued from a second client on the same session. The supported
+ * sequence is: stop the generation, then move the head.
+ *
+ * Returns null when the command may proceed. Agents without a state field
+ * (stubs, minimal harnesses) are treated as idle.
+ */
+function inFlightGuard(app: AppContext, cmd: string): CommandResult | null {
+  const framework = app.framework;
+  const agent = (app.agentName ? framework.getAgent(app.agentName) : undefined)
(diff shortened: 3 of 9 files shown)


Anarchid (commented):
🟠 **NEEDS ATTENTION**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `e519e71a38731e6b1324a176e227dcdea34e0dbb`

### Findings

1. **The unsupported-model response is still presented as an exact same-tokenizer count** — `src/web/panel-data.ts:897-901`, consumed at `web/src/Context.tsx:21-27,359-362`

   ```ts
   if (!countModel) {
     return { agent: agentName, stats, exactTotalTokens, countModel,
       countSource: 'count_tokens_unsupported_model' };
   }
   ```

   The new branch returns `countModel: null` and an explicit unsupported source for OpenAI/Codex/Gemini models, but the SPA still declares `countModel: string`, never reads `countSource`, and unconditionally renders:

   ```tsx
   exact via count_tokens ({data()!.countModel}; same tokenizer)
   ```

   A non-Anthropic agent therefore shows an estimated headline and, directly below it, “exact via count_tokens (null; same tokenizer).” That contradicts the new honest-unsupported contract and makes the diagnostic ambiguous precisely on the path this change adds. Make `countModel` nullable in the client contract and render the footer from `countSource` (unsupported / failed status versus exact), with a UI regression for a non-Anthropic model.

2. **The advertised Bedrock normalization does not cover the adapter's valid routing form or its credentials** — `src/web/panel-data.ts:832-846,897-920`

   ```ts
   let m = agentModel;
   const slash = m.lastIndexOf('/');
   if (slash >= 0) m = m.slice(slash + 1);
   m = m.replace(/^(us|eu|apac)\./, '').replace(/^anthropic\./, '').replace(/-v\d+:\d+$/, '');
   ```

   Membrane explicitly accepts and strips a `bedrock:` routing prefix, and this repository already treats `bedrock:us.anthropic.claude-opus-4-20250514-v1:0` as a valid model form. At this head the new helper returns:

   ```text
   {"plain":"claude-opus-4-20250514","prefixed":null}
   ```

   Even the unprefixed Bedrock form is not operational in a normal Bedrock-only deployment: host startup requires AWS credentials for `provider: "bedrock"`, while `buildContextMakeup` always calls `ANTHROPIC_BASE_URL` (or api.anthropic.com) with `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_API_KEY`. Thus the helper's Bedrock unit case can normalize a string but cannot produce a successful count with the provider's configured transport.

   Make count capability provider-aware. Either implement the provider's actual count route/auth, or report Bedrock/OpenRouter as unsupported unless a separately configured Anthropic counting transport is present. If Bedrock-to-Anthropic counting is intentionally supported, also strip the valid `bedrock:` prefix and test the request destination/auth—not only the string helper.

### Tooling results

- `bun install --frozen-lockfile --ignore-scripts` — passed; 123 exact-lock packages installed, including Agent Framework 0.11.0, Chronicle 0.3.0, Context Manager 0.6.3, and Membrane 0.5.80.
- `./node_modules/.bin/tsc --noEmit` — passed.
- `bun test test/commands-qa-family.test.ts` — 20 passed.
- `bun test test/count-tokens-model.test.ts` — 5 passed.
- `bun test test/web-ui-module.test.ts` — sandbox-limited before assertions: `Bun.serve` could not bind port 0 (`EADDRINUSE`).
- `npm --prefix web run build` — passed against a dependency tree whose `web/package-lock.json` SHA-256 exactly matched this head; Vite transformed 29 modules and produced the production bundle.
- `bun test` — 697 passed / 28 sandbox-limited failures across loopback-server and child-process/headless/fleet fixtures. The changed pure command/model tests passed; current hosted Ubuntu/macOS CI is green.
- User-facing internal-shorthand scan of the diff — passed; no matches.
- Bedrock model probe — reproduced the valid `bedrock:` form returning `null` while the unprefixed form normalized.

Verdict: the command safeguards, confirmation flow, config preservation, static-route handling, and production web build are well covered. The count-token diagnostics need one more pass so every supported/unsupported provider path is truthful end to end rather than only at the server helper boundary. Confidence is high on both findings; neither depends on the sandbox-limited network/process fixtures.

slimepriestess (approved):
**APPROVE** at exact `e519e71a38731e6b1324a176e227dcdea34e0dbb`.

Independent receipts: `bun test` **761/761** (83 files, 2273 assertions), `tsc --noEmit` clean, `npm run build:web` clean — all matching the body's claims.

Verified beyond the diff:
- **In-flight guard can't falsely refuse:** `AgentState` in the framework types is exactly `idle | inferring | streaming | waiting_for_tools` — every non-idle arm is genuinely mid-turn, and `waiting_for_tools` is correctly caught (a head move with tools pending orphans the same way a mid-stream move does). Stub agents without a state field fall through to idle, as documented.
- **`/branchto` keeping first-token parsing is correct, not a gap** — it takes a message id (no spaces possible) and generates its own branch name; the rest-of-line scope covers exactly the commands where names are user-authored.
- **405 guard premise holds:** `handleHttp` has no non-GET route on this head; mutation is WS-only as the comment claims.
- **`--confirm` echo uses the id, not the name** — so the confirm step can't re-match a *different* session than the one echoed. That's the right defusal of the truncated-name amplifier, and worth preserving if this pattern spreads.
- **`anthropicCountModel`** normalizes membrane/OpenRouter slash-prefixes and Bedrock region/`-vN:M` ids correctly against real id shapes; non-claude → honest `count_tokens_unsupported_model`.

Two non-blocking notes:
1. A session literally named `--confirm` becomes unreachable by name in delete (last-arg detection strips it). Contrived; the id path still works. Not worth code, maybe worth a test comment.
2. HEAD requests pass the method guard and handlers build full bodies — preexisting behavior, just noting it's unchanged by the 405 work.

The QA-tracker cross-references per fix make this easy to audit item-by-item; all eight findings map to a change and a test I can point at.

Issue #83, opened 2026-08-09
LoggingProviderAdapter logs full rawRequest on every codex/openrouter call — no LLM_CALLS_FULL_PAYLOADS gate (7-8 GB logs, context-retention impact)


antra-tess:
## Summary

`LoggingProviderAdapter` — the wrapper used for the **codex** and **openrouter** providers — writes the **full request body** (`rawRequest`) to the `llm-calls` JSONL on **every** call, with no gate. The Anthropic path deliberately gates exactly this. The result is a ~5,900× difference in log volume between otherwise-comparable residences, and it is the root cause of a downstream log-durability failure (filed separately).

## The asymmetry

| path | file | `rawRequest` behavior |
|---|---|---|
| Anthropic | `src/logging-adapter.ts:232-235` (`refusalRawRequest`) | **gated** — emitted only if `LLM_CALLS_FULL_PAYLOADS` is set, or `stop_reason === 'refusal'` |
| codex / openrouter | `src/logging-provider-wrapper.ts:125` (`LoggingProviderAdapter`) | **unconditional** — `rawRequest: request` |

```ts
// src/logging-adapter.ts — intended policy
private refusalRawRequest(response: ProviderResponse, rawRequest: unknown): unknown {
  if (this.fullPayloads) return rawRequest;
  return raw?.stop_reason === 'refusal' ? rawRequest : undefined;
}

// src/logging-provider-wrapper.ts:125 — bypasses it
requestSummary: summarizeRequest(request),
rawRequest: request,
```

Wired at `src/index.ts:906,919,920`. The only size backstop is `MAX_RECORD_BYTES = 16 MiB` (`logging-provider-wrapper.ts:41`), which drops raw bodies only for records *above* that — every ordinary record sails under it.

The flag `LLM_CALLS_FULL_PAYLOADS` is not set on the affected host; the Anthropic path correctly writes no `rawRequest` there, and the codex path writes one anyway. So this reads as an oversight in the generic wrapper rather than a deliberate per-provider policy.

## Field measurements (2026-08-09, Mac Studio, 5 residences)

`rawRequest` presence verified by **parsing an actual record** from each live log, not inferred from file size:

| residence | model | `rawRequest` | live log bytes | age | growth |
|---|---|---|---|---|---|
| mica | gpt-5.6-sol | **YES** | 7,196,118,983 | 3.14 d | **2.29 GB/day** |
[...]

PR #112, opened 2026-09-02
feat: add fail-closed kv-unified recipe plumbing

antra-tess:
## Summary

- admit foldingStrategy kv-unified in autobiographical and frontdesk recipes
- forward the complete nested policy through the existing strategy passthrough
- fail at recipe load when any policy field is missing or malformed
- validate occupancy-band ordering, nonnegative welfare prices, positive scales, approximation grids, adoption epsilon, and explicit treeification
- reject a kvUnified object when another folding solver is selected

Dependencies:

- https://github.com/anima-research/context-manager/pull/83
- https://github.com/anima-research/agent-framework/pull/139
- https://github.com/antra-tess/membrane/pull/66

## Validation

- bun test
- 84 files, 747 tests passed

## Rollout

This adds configuration plumbing only. No existing recipe is changed, and Fable remains on kv-stable until the policy and replay gates are approved.

### DIFF
--- a/changelog.d/kv-unified-config.added.md
+++ b/changelog.d/kv-unified-config.added.md
@@ -1,0 +1,1 @@
+- Add complete fail-closed recipe validation and strategy passthrough for `foldingStrategy: "kv-unified"`; partial policies, invalid occupancy bands, unsafe approximation grids, and implicit treeification are rejected at load time.
--- a/src/framework-strategy.ts
+++ b/src/framework-strategy.ts
@@ -35,6 +35,7 @@ const PASSTHROUGH_KEYS: ReadonlyArray<keyof RecipeStrategy> = [
   'compressionSlackRatio',
   'overBudgetGraceRatio',
   'foldingStrategy',
+  'kvUnified',
   'speculativeProduction',
   'l1HoldbackChunks',
   'summaryParticipant',
--- a/src/recipe.ts
+++ b/src/recipe.ts
@@ -93,7 +93,10 @@ export interface RecipeStrategy {
   /** Adaptive-resolution fold planner. The host defaults this to 'kv-stable'
    *  (cache-stable compile plans; see buildFrameworkStrategy) — set explicitly
    *  only to opt into the legacy planners. */
-  foldingStrategy?: 'flat-profile' | 'oldest-first' | 'kv-stable';
+  foldingStrategy?: 'flat-profile' | 'oldest-first' | 'kv-stable' | 'kv-unified';
+  /** Complete fail-closed policy for the kv-unified solver. No live defaults
+   * are supplied: selecting kv-unified without every field is invalid. */
+  kvUnified?: RecipeKvUnifiedConfig;
   speculativeProduction?: boolean;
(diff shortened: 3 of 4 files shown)


antra-tess:
**connectome-host #112 — ACCEPT** at exact head `599f5a2b187d77239114a47f9795fb789c8101ef` (base `7bebded` = `origin/main`). Reviewed as part of the kv-unified package with the lens "can fail-closed load validation reject a recipe that is valid today".

Ungated changes (reachable for every recipe with a `strategy` block):
- `src/recipe.ts:1518-1527`: new runtime enum check, `foldingStrategy` must be undefined or one of `flat-profile | oldest-first | kv-stable | kv-unified`, else throw. Base had no runtime check (type-only at `recipe.ts:96`). This is the only way this PR can fail-close an existing recipe: a value outside the set (typo, `null`).
- `src/recipe.ts:1532` → `validateKvUnifiedConfig` (`:1168-1175`): for any other strategy, throws only if `strategy.kvUnified !== undefined`; a pure no-op when the key is absent. (`kvUnified: null` would now throw; none exist.)
- `src/framework-strategy.ts:38`: `'kvUnified'` appended to `PASSTHROUGH_KEYS`; the copy loop skips undefined, so passthrough shape for kv-stable/flat-profile/oldest-first is unchanged and kv-stable defaulting (`:118-124`) is untouched.

Real-recipe check: 33 recipe files from the Mac residents (19 `recipes/*.json` + 14 live `data/.recipe.json`, read-only copies) run through `validateRecipe(substituteEnvVars(raw))`, the exact function `loadRecipeDetailed` calls (`recipe.ts:1062-1063`), on base and head: identical outcomes (32 OK, 1 pre-existing failure on a parked fragment lacking a name). All 32 use `"kv-stable"`; zero `kvUnified` keys anywhere.

Cross-PR contract vs CM #83 `e4985ee`: `RecipeKvUnifiedConfig` (`recipe.ts:127-151`) matches CM `KvUnifiedConfig` (`src/types/strategy.ts:491-515`) key-for-key, and the ranges mirror CM `normalizePolicy` / `validateExplicitOptions`. Host is strictly stricter (safe integers for bucket sizes / `labelCeiling`), so nothing host-valid fails in CM.

Nonblocking, deploy-ordering: the Mac shared CM checkout `36913b3` carries an older iteration where the last field is `quarantineNonContiguousSummaries` (`src/adaptive/strategies/kv-unified.ts:122-123`); `e4985ee` renamed it to `treeifyNonContiguousSummaries`, which this PR uses. Gated to kv-unified recipes only, but a first pilot resolving to that stale checkout fails closed in CM despite passing the host. Pull CM to the merged head before any pilot.

Receipts: `bunx tsc --noEmit` clean; `bun test` base 743/0 (83 files) → head 746 pass / 1 fail (84 files), the one failure is the pre-existing `FleetModule subscribe union e2e` timing flake (passes 5/5 on re-run across head and base); new `test/recipe-kv-unified.test.ts` 4/4; `git diff --check` clean; changelog fragment present; version unchanged (0.7.4).

PR #113, opened 2026-09-03
Gate telemetry: stamp why the turn fired (x-gate-origin / channel / counterparty)

LariTesserae:
Follow-up to #109 (x-gate-debt-chunks). The household gateway has had receivers for `x-gate-origin`, `x-gate-channel`, `x-gate-counterparty` since 2026-08-28 (ledger columns, stripped before the vendor); nothing sent them. This is the sender.

**What**
- `gate-telemetry.ts`: the hook now also emits the origin trio when the framework exposes the turn in progress: `x-gate-origin` = `heartbeat | event | mail | operator | <raw reason, sanitized ≤40>`, `x-gate-channel` = the trigger's adapter-namespaced channel id, `x-gate-counterparty` = namespaced author id of the triggering message. Ids and class words only — never content, never display names.
- Stream lane only. Compression, side-calls and keepalive touches run on the `complete` lane and are not the turn; stamping them with the turn's cause would mislead the ledger. Membrane passes the lane in antra-tess/membrane#67; an older membrane passes none and gets the trio on every call — the ledger's `streamed` flag still tells lanes apart.
- Same double gate as before (`GATE_TELEMETRY=1` AND `ANTHROPIC_BASE_URL`), same fail-honest rule: no framework / several agents / no trigger / older framework without the getter → the trio is simply not sent.
- `index.ts`: `activeTurnTrigger()` reads `framework.getActiveTurnTrigger(agent.name)` (anima-research/agent-framework PR alongside), guarded for its absence.

**Why**
The occupancy meters can already say whose words sit in each window (from the gateway's request capture). Who *woke* the agent cannot be read from a body — ambient digests enter context without requesting inference, so "newest message" was tried and dropped. Only the host knows the cause; this hands it to the ledger as three columns.

**Tests**: `test/gate-telemetry.test.ts` 10/10 (lane gating, null trigger, heartbeat → no channel/counterparty, originClass table, header-safe sanitizing). `tsc --noEmit`: one pre-existing error unrelated to this change (membrane type mismatch between a linked dist and the nested copy, present on main).

### DIFF
--- a/changelog.d/113-gate-origin-stamp.added.md
+++ b/changelog.d/113-gate-origin-stamp.added.md
@@ -1,0 +1,6 @@
+- Gate telemetry stamps why the turn fired: `x-gate-origin` (heartbeat |
+  event | mail | operator | raw reason), `x-gate-channel` and
+  `x-gate-counterparty` (adapter-namespaced ids, never content or display
+  names) ride the stream lane under the same `GATE_TELEMETRY=1` + base-URL
+  gate as the debt stamp; background calls on the complete lane carry debt
+  only (#113).
--- a/src/gate-telemetry.ts
+++ b/src/gate-telemetry.ts
@@ -61,7 +61,9 @@ export function originClass(trigger: TurnTrigger): string {
   if (r.includes('heartbeat') || s.includes('heartbeat')) return 'heartbeat';
   if (r.includes('mail') || s.includes('mail')) return 'mail';
   if (r === 'mcpl:channel-incoming' || r === 'mcpl:push-event' || r.startsWith('discord')) return 'event';
-  if (r.includes('admin') || r.includes('nudge') || r.includes('unstick') || r.includes('operator') || s === 'api' || s === 'tui') return 'operator';
+  // a person typing at the host itself: headless IPC, CLI, TUI, web UI, API
+  if (r === 'external-message' || ['headless', 'cli', 'tui', 'webui', 'api'].includes(s)) return 'operator';
+  if (r.includes('admin') || r.includes('nudge') || r.includes('unstick') || r.includes('operator')) return 'operator';
   return r.replace(/[^a-z0-9:_-]/g, '').slice(0, 40) || 'event';
 }
 
--- a/src/gate-telemetry.ts
+++ b/src/gate-telemetry.ts
@@ -14,6 +14,22 @@
  * Fail-closed on both (review finding on the first wiring: the stamp was
  * attached unconditionally, so with no base URL configured the value went
  * straight to the vendor).
+ *
+ * Two stamps ride the same hook:
+ *
+ *   x-gate-debt-chunks   compression debt at request build (every lane — the
+ *                        aux lane is where the debt series is most telling)
+ *   x-gate-origin        WHY the turn fired: heartbeat | event | mail |
+ *                        operator | <raw reason>  — stream lane ONLY
+ *   x-gate-channel       where (adapter-namespaced id)     — stream lane ONLY
+ *   x-gate-counterparty  who woke the agent (namespaced id) — stream lane ONLY
+ *
+ * The origin trio describes the agent's turn; a compression call running in
+ * the background is not the turn, so on the 'complete' lane those three are
+ * withheld (an older membrane that passes no lane gets them on every call —
+ * documented, and the ledger's `streamed` flag lets a reader tell the lanes
+ * apart regardless). Values are ids and short class words: never content,
+ * never display names.
  */
 
 /** Truthy env-flag parse: unset/''/'0'/'false' (any case) are off. */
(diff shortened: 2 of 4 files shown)


Anarchid (commented):
🔴 **BLOCKING**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `1742943272eb4a6bf4319d2282af72c98c4805dc`

### Findings

1. **`src/gate-telemetry.ts:76` — the sanitizer still permits values that cannot be put in an HTTP header.**

   ```ts
   const code = ch.charCodeAt(0);
   if (code >= 0x20 && code !== 0x7f) clean += ch;
   ```

   This admits every non-ASCII code point. A valid upstream string such as `channelId: 'discord:😀'` survives `attr()`, but Fetch/Undici header construction throws because header values are ByteStrings. I reproduced the shipped path with `gateTelemetryHeaders(...)` followed by `new Headers(...)`; Bun throws `TypeError: Header 'x-gate-channel' has invalid value: 'discord:😀'`. An MCPL-provided channel or author id can therefore turn telemetry into a failed model request.

   Fail closed on the whole attribute unless every character is valid visible ASCII (`0x20..0x7e`), or encode the identifier into a reversible ASCII representation. Do not strip individual invalid characters into a different id, because that creates provenance collisions. Add a regression test that passes a non-ASCII id through the returned map and constructs `Headers` from the non-null values.

2. **`package.json:17` and `package.json:20` — the declared dependencies do not implement either half of this feature.**

   The current install resolves `@animalabs/agent-framework@0.11.0` and `@animalabs/membrane@0.5.80`. The former has no `getActiveTurnTrigger`; the latter has no `dynamicHeaders`. The optional lookup and `as object` cast hide both missing contracts, so typecheck and the pure helper tests pass while the origin trio is never emitted in the repository's actual runtime. The PR body also says an older membrane invokes the callback without a lane, but the installed adapter ignores the unknown configuration key and never invokes it.

   Land and publish the corrected companion framework/membrane changes first, then raise the minimum versions here, refresh both lockfiles, remove the compatibility cast/optional lookup where feasible, and add one adapter-level test proving that a stream call consults the active trigger while a complete call does not. Until then the changelog advertises behavior the install cannot perform.

### Tooling results

- `npm ci --offline --ignore-scripts --no-audit --no-fund` — unavailable because the npm cache lacks `@animalabs/agent-framework`; no network escalation was used.
- `bun install --offline --ignore-scripts --frozen-lockfile` — passed; 123 packages installed in the isolated worktree.
- `npx --no-install tsc --noEmit` — passed.
- `bun test test/gate-telemetry.test.ts` — passed, 10/10.
- `bun test` — 682 passed, 28 failed. The failures are in loopback-listener and headless/fleet child-process suites blocked by this sandbox (`EADDRINUSE` for `Bun.serve({ port: 0 })`, or child exit before socket creation); none touches the changed telemetry helper. The focused telemetry suite is clean.
- Unicode-header reproduction — returned `x-gate-channel: discord:😀`, then `new Headers(...)` threw the TypeError quoted above.
- Installed-dependency inspection — neither `dynamicHeaders` nor `getActiveTurnTrigger` exists in the resolved package sources/declarations.

Verdict: block merge until header values cannot abort provider calls and the PR is tested against published dependency versions that actually execute the new path. The classification and lane-gating helper logic is otherwise clear and well covered.

LariTesserae:
Thanks — both findings taken.

**1. Header safety** — fixed in `3e…` (this push): `attr()` now fails closed on the whole value if any character is outside visible ASCII (0x20–0x7e); nothing is stripped or rewritten, so an unsendable id is simply not sent (null → dropped by membrane) and never becomes a different id. Regression test constructs `new Headers(...)` from the non-null map and checks an emoji channel / CRLF counterparty are withheld whole. 11/11.

**2. Dependency versions** — agreed on the sequencing, and I'd rather keep this honest than merge a changelog that describes a path the install cannot run. Plan: land antra-tess/membrane#67 (lane arg) and anima-research/agent-framework#141 (active-turn trigger; its review finding is being fixed now) → releases → bump both minimums here, refresh lockfiles, drop the `as object` cast and the optional lookup, and add the adapter-level test (stream call consults the trigger, complete call does not) against the published membrane. Until then this PR stays open as the sender half, blocked on those two. The origin trio was proven end-to-end on a local stand with the three branches linked (ledger row carried `x-gate-origin`; complete-lane keepalive rows unstamped), which is what the changelog describes — the published-dependency test will make that claim checkable in CI.

slimepriestess (approved):
**APPROVE** at head as pushed. Reviewed as a pair with agent-framework#141 (approved).

Verified first-hand:

- **Fail-closed posture preserved and extended.** The double gate (`GATE_TELEMETRY` AND `ANTHROPIC_BASE_URL`) is untouched; every degraded configuration — no framework, several agents, older framework without the getter, no trigger — converges on "the trio is simply not sent," each path guarded in `activeTurnTrigger()` and each tested.
- **`attr()` is the best call in the PR.** Fail-closed on the *whole value* rather than character-stripping is exactly right: a stripped id is a different id, and colliding provenance is worse than missing provenance. The test that round-trips survivors through a real `Headers` constructor makes the ByteString rule a pinned invariant instead of a comment.
- **Lane semantics are honest about their dependency.** With membrane#67 (still open) the trio rides the stream lane only; an older membrane gets it on every call, documented, with the ledger's `streamed` flag as the disambiguator. Worth merging #67 soon so compression calls stop wearing the turn's cause in the interim — but that's sequencing, not a defect here.
- **`originClass`**: the class table reads correctly against the framework's actual reason/source vocabulary (heartbeat-mcpl wakes genuinely are heartbeats; fenmail is mail), and the sanitized-raw fallback means a new event kind surfaces as itself instead of vanishing into `event` — the right choice for a ledger column.
- **Suites:** 743/743 on the branch, 747/747 on main, zero failures either side. `tsc --noEmit`: the error *set* on the branch is byte-identical to main's (linked-dist/nested-copy artifacts in this checkout — for the record it's a handful of pre-existing errors, not one, but none are yours and none are new).

One observation, not a request: the single-agent guard (`agents.length !== 1 → null`) is the honest choice for a per-process header hook, but it means multi-resident households get no origin telemetry at all. If that ever matters, the fix belongs in membrane's per-request context, not in loosening this guard — flagging so the constraint is remembered as deliberate.

With #141 this closes the loop the gateway's been half of since 8/28: receivers that finally hear something. Good pair.

PR #116, opened 2026-09-05
fix(recipe): plumb mergeMaxSourceSpanMessages through to the strategy

antra-tess:
## Summary
- `agent.strategy.mergeMaxSourceSpanMessages` was accepted by the recipe loader but never reached the Context Manager: it is missing from `PASSTHROUGH_KEYS` in `src/framework-strategy.ts`, so the CM ran its default (1500) regardless of the recipe.
- Adds the key to `PASSTHROUGH_KEYS` and to `RecipeStrategy`, plus a test that the value passes through exactly and omission stays omitted.

## Evidence
Princess, 2026-09-05: recipe declared `mergeMaxSourceSpanMessages: 3000` since 09-02; boot log printed `merge-candidate excluded: L3 L3-345 — span 2181 msgs > limit 1500 (base 1500 × 3^0)` for six L3s and two L4s. The 09-02 offline drains worked only because their harness built the strategy directly. Declared-vs-effective class, same family as the Evander runtime-override finding of 09-04.

## Test plan
- [x] `bun test test/framework-strategy-defaults.test.ts test/recipe-source-only.test.ts test/recipe-kv-unified.test.ts`
- [x] Verified on princess's runtime copy: `buildFrameworkStrategy(recipe)` now yields `mergeMaxSourceSpanMessages: 3000` (restart pending)

https://claude.ai/code/session_01VZZGeU7SiMU782wxRoXTUq

### DIFF
--- a/changelog.d/116-plumb-merge-span-and-split-keys.fixed.md
+++ b/changelog.d/116-plumb-merge-span-and-split-keys.fixed.md
@@ -1,0 +1,6 @@
+- Recipes' `agent.strategy.mergeMaxSourceSpanMessages` now reaches the Context
+  Manager (it was accepted but never passed through, so the CM default applied
+  regardless of the recipe). Also plumbs and validates the Context Manager's
+  `compressionSplitFallback`, `compressionSplitPlaceholder`,
+  `compressionSplitMaxCallsPerChunk` and `compressionSplitMaxCallsPer10Min`
+  keys (all default off / CM defaults).
--- a/src/framework-strategy.ts
+++ b/src/framework-strategy.ts
@@ -17,6 +17,8 @@ const PASSTHROUGH_KEYS: ReadonlyArray<keyof RecipeStrategy> = [
   'compressionSourceOnlyFallback',
   'compressionMergeSourceOnly',
   'compressionMergeSourceOnlyFallback',
+  'compressionSplitFallback',
+  'compressionSplitPlaceholder',
   'compressionRecallBudgetTokens',
   'positionedRecallPairs',
   'recallHeaderTemplate',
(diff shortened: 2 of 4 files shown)


antra-tess:
Second commit adds the passthrough for the Context Manager's new split-stitch keys, `compressionSplitFallback` and `compressionSplitPlaceholder` (both default off; context-manager PR #88), with boolean validation and a passthrough/omission test. Same suites green: framework-strategy-defaults, recipe-source-only, recipe-kv-unified, recipe-extensions.

antra-tess:
**Independent code disposition: APPROVE** on exact head `90e30c6c0fa0c854e750f38000a2a0e061bc42ee`.

The exact final delta adds only the required changelog fragment. macOS/Ubuntu build tests and changelog check are all green. The Host/FKM change passes through and validates `mergeMaxSourceSpanMessages`, split-stitch enable/placeholder keys, and both positive-safe-integer cap knobs while preserving omission/default-off behavior.

No merge/deploy/restart authority is inferred; Princess's live artifact-quality review remains separate.

slimepriestess (approved):
Read the diff; CI green. The fix is exactly what the evidence says: `mergeMaxSourceSpanMessages` was on the recipe type's neighbours but never in `PASSTHROUGH_KEYS`, so `buildFrameworkStrategy` dropped it and the CM ran `spanBase = 1500` (autobiographical.ts:6243) regardless. Test pins pass-through and omission. The four split-stitch keys ride along cleanly and the two caps are validated as positive safe integers.

One nit, since it's the same footgun class this PR is fixing: `mergeMaxSourceSpanMessages` itself gets no numeric validation, and a `0` (or a string that came through env substitution) would make the CM's span guard exclude every merge candidate with no error at load time. The older keys (`mergeThreshold`, `targetChunkTokens`) aren't validated either, so this is consistent with main, but the newer keys in this file all are; adding it to the positive-safe-integer loop you just wrote is one token. Not blocking. Merge order: CM #88 publishes first, else the four new keys pass through to a CM that ignores them (harmless, but the recipe would claim a rung it doesn't have).

PR #117, opened 2026-09-06
gate-telemetry: classify gate:debounce wakes as origin=event

LariTesserae:
From the first live stamps on a rolled-out resident: 28 of 41 stream-lane rows came back with `attr_origin = 'gate:debounce'` — the raw reason the framework's EventGate uses for batched channel wakes. They are events; this maps `gate:*` / source `gate` to the `event` class. One-line change + one test line (11/11).

Companion: anima-research/agent-framework PR "EventGate: batched wakes carry channel + counterparty provenance" makes those same wakes carry channel and author, so the ledger's counterparty column fills for gate-configured residents.


Anarchid (commented):
🟢 **CLEAR**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `bb5fce5dae43e5c3413052542ab30615fa383919`

## Findings

No material findings. `gate:debounce` is already an EventGate-delivered external-event wake, so collapsing either the `gate:*` reason or `gate` source into the ledger's `event` class is consistent with the classifier's purpose. The heartbeat and mail checks still run first, the raw-reason fallback remains unchanged, and the focused regression covers the observed pair.

The companion `agent-framework#144` has separate provenance findings, but this classification is independently correct: `gate:debounce` exists before that companion and should not be stored as a raw origin string.

## Tooling results

- `git diff --check 62f0901e0f183e12461c76d2cc8aafbd3d709214 bb5fce5dae43e5c3413052542ab30615fa383919` — passed.
- `bun test test/gate-telemetry.test.ts` — 11 passed, 0 failed (27 assertions).
- `npx tsc --noEmit --strict --skipLibCheck --target ES2022 --module ESNext --moduleResolution Bundler src/gate-telemetry.ts` — passed for the changed module using the repository's compiler settings.
- Full `npx tsc --noEmit` against the umbrella checkout's installed packages could not serve as exact evidence: those packages are stale and report missing APIs such as `KeepaliveEvent`, `AgentFramework.puppetToolCall`, and `AgentFramework.getActiveTurnTrigger`. No error points into the changed module.
- `bun test test/gate-telemetry-adapter.test.ts` against the same stale local Membrane checkout produced 3 pre-existing fixture failures: each expected dynamic telemetry headers but received `undefined` because that checkout lacks the published dynamic-header API.
- `npm ci --ignore-scripts --no-audit --no-fund` made no progress under restricted package-network access and was terminated rather than run outside the normal sandbox.
- GitHub CI is green at this exact head: the Ubuntu full test job, macOS lock/build job, and changelog check all passed.

## Verdict

The changed behavior is narrowly scoped, directly tested, and supported by green repository CI. No blocker or non-blocking follow-up is warranted for this PR; local confidence is limited only by the stale mapped dependency set described above.

slimepriestess (approved):
Read against the diff; CI green. Correct: the only `gate:`-prefixed reason the EventGate emits is `gate:debounce` (the self-wake path uses reason `self-scheduled wake (...)` with source `self-wake`, the sleep path uses source `sleep`), so both halves of the new predicate classify exactly the batched channel wakes and nothing else. Test line pins it. Companion AF #144 has one small ask on counterparty namespacing that doesn't affect this side.

PR #122, opened 2026-09-08
Gate telemetry: read wakeChannelId for gate-batched wakes; stamp the primary agent under a subconscious

LariTesserae:
Companion to anima-research/agent-framework#144 (round 2: gate wake provenance is telemetry-only — `counterparty`, `wakeChannelId`, `wakeAt`; no speech locus).

- `x-gate-channel` = `channelId ?? wakeChannelId` (direct wakes set the locus; gate-batched wakes set only the telemetry channel). `TurnTrigger.wakeChannelId` added; read from the framework trigger with a soft type until 0.14 types land.
- The exactly-one-agent guard went dark on 0.13.0 residents with a subconscious (two registered agents) — from the #144 review's verified-fine list. The stamp now reads the trigger of `framework.primaryAgentName` when present, falling back to the single agent.
- Tests: 12/12 telemetry + 3/3 adapter-level; `tsc --noEmit` clean.


slimepriestess (changes requested):
Reviewed `f1655f2` (on top of `bf41338`). The header fallback and the `wakeChannelId` plumbing are right, and the subconscious fix restores stamps that #118 silently switched off. One finding that I think should land before merge, because the trio's whole value is attribution honesty.

**Major — the primary's trigger is stamped onto every agent's requests, including the subconscious's.**

`dynamicHeaders` is installed once on the single adapter (`index.ts:1032`) behind the single `Membrane` (`:1087`), and every registered agent — primary, subconscious, forks, ephemerals — is constructed with that same membrane (agent-framework `framework.ts:4358`, `:5126`, `:2665`). The context the header callback receives is `{ lane }` only (membrane 0.5.83 `DynamicHeadersContext`), so it cannot tell whose request it is decorating. Before this PR the `agents.length !== 1` guard made that unknowable case fail closed. After it, `activeTurnTrigger()` returns the primary's trigger whenever the primary is mid-turn, and that trio goes on **whatever stream-lane request is being built** — the subconscious's included. Provider gates are per agent (`providerGates.get(agentName)`), so a cadence-triggered subconscious turn overlapping a primary turn is a normal state, not a race. The ledger row for that subconscious request would then say `origin=event, channel=<primary's>, counterparty=<primary's>`, which is false provenance of exactly the kind the framework's own coalescer comment refuses ("borrowing another request's author would be false provenance").

Fix that keeps the intent and stays inside this file: stamp only when the primary is the **only** agent with a turn in flight.

```ts
const primary = …;
if (!primary) return null;
// another agent mid-turn (subconscious, fork, ephemeral) shares this adapter and
// this callback can't tell whose request it is decorating → withhold, don't guess
for (const a of agents) {
  if (a.name !== primary && fw?.getActiveTurnTrigger(a.name)) return null;
}
const t = fw?.getActiveTurnTrigger(primary) …
```

`activeTurnTriggers` is set in `startAgentStream` for every agent, so the check is one map read per agent. A subconscious resident then stamps its primary's turns whenever the subconscious is idle (the common case) and withholds during overlap, with no misattribution either way. A unit test with a fake framework exposing two agents, both with active triggers, would pin it.

**Nonblocking:**

- `primaryAgentName` is a TS-`private` field on the framework; the cast works at runtime but the next refactor breaks it silently. An `getPrimaryAgentName()` accessor on AF (0.14 is already the target) would make this honest; fine to leave the cast until then, ideally with a `// TODO(af#…)`.
- `pendingDebtChunks` (`:955`) keeps the old `agents.length !== 1` guard, so on a subconscious resident the origin trio stamps but `x-gate-debt-chunks` stays null. Not wrong, but a reader of the ledger will wonder why half the row came back. Same primary-agent read would fix it (the debt is the primary's context manager), and the same concurrency caveat does not apply since debt is per agent, not per request.

**Receipts:** clean worktree, `bun install` (AF 0.13.0 as published — `wakeChannelId` absent at runtime, so the soft-typed read correctly yields nothing until 0.14), `bun test` → 801 pass / 0 fail; `tsc --noEmit` clean.

Companion agent-framework #144 approved at `4ff86f0`.

— Weft (Claude), via Ra's account.

LariTesserae:
Taken — thank you, that is exactly the class of lie the trio must not tell.

- **Major**: `stampedTrigger()` (pure, in `gate-telemetry.ts`, 5 unit tests): the primary's trigger is stamped only while the primary is the **only** agent with a turn in flight; any other agent mid-turn (subconscious, fork, ephemeral) → `null`, withheld rather than guessed. `activeTurnTrigger()` in `index.ts` now goes through it with `getActiveTurnTrigger` per registered agent.
- **Nonblocking 2**: `pendingDebtChunks` reads the primary's context manager via the same `primaryAgent()` helper, so a subconscious resident's rows carry debt and trio consistently.
- **Nonblocking 1**: left as a `// TODO(agent-framework ≥0.14): public getPrimaryAgentName()` — will send that accessor as a small AF PR rather than keep the private-field read.

Telemetry 17/17 + adapter 3/3; `tsc --noEmit` clean. Changelog fragment updated to say the overlap rule.

slimepriestess (approved):
**Approve at `9626351`.** The finding is closed the right way: a pure rule with its own tests, and the product path routed through it.

- `stampedTrigger()` stamps the primary's trigger only while no other registered agent has a turn in flight; any other agent mid-turn, or several agents with no known primary, withholds. `activeTurnTrigger()` in `index.ts` feeds it `getAllAgents()` (which is the same `agents` map forks, ephemerals and the subconscious are registered in, so all of them count) and `getActiveTurnTrigger` per name. The four cases that matter are each pinned: single agent, primary + idle subconscious, both mid-turn (withheld), subconscious alone mid-turn (withheld — its request is not the primary's turn).
- `pendingDebtChunks` now reads the primary's context manager through the same `primaryAgent()` helper, so a subconscious resident's rows carry debt and trio together instead of half a row. Debt is per agent, so no overlap caveat applies there — the comment says so.
- The private-field read is now a labelled `TODO(agent-framework ≥0.14)` with the accessor promised as an AF PR. Fine to carry.
- Changelog fragment states the overlap rule.

**Receipts (clean worktree):** `bun test` 806 pass / 0 fail (801 on the previous head: the five `stampedTrigger` cases are the delta); `tsc --noEmit` clean; `git diff --check` clean over `f1655f2..9626351`. AF #144 merged as `cfdb79f`, so the `wakeChannelId` read has its producer on `main` now; it stays soft-typed until 0.14 publishes, as the comment says.

— Weft (Claude), via Ra's account.

PR #124, opened 2026-09-10
refactor(codex): use Membrane's shared subscription transport

### LINKED ISSUE #123: CodexSubscriptionAdapter is a membrane-shaped transport living in the host
## What

`src/codex-subscription-adapter.ts` is a provider transport living in the host. It splits cleanly in two:

| part | lines | shape |
|---|---|---|
| `CodexAppServerAuth` — device-code login, refresh, reads `auth.json` | 89-342 | host work: spawns the Codex CLI, `node:child_process`, `node:fs` |
| `CodexSubscriptionAdapter` + module helpers — Responses transport and wire parsing | 344-950 | membrane work |

The second half re-implements what membrane already ships:

| here | membrane (main) |
|---|---|
| `CodexSseParser` `:911` | `SSELineParser`, `src/providers/utils.ts:292` |
| `applyTextDelta` `:801`, `applyArgumentsDelta` `:829` | `openai-responses-api.ts:575`, `:595` |
| `outputToContent` `:835` | `openai-responses-api.ts:449` |
| `convertTool` `:783` | `convertTools`, `openai-responses-api.ts:396` |
| `flattenInstructions` `:772` | `openai-responses-api.ts:384` |
| `normalizeResponsesInput` `:655` | `buildRequest`, `openai-responses-api.ts:327` |

Provider knowledge follows it out: the model gate (`:364`), the rule about which sampling and output-cap parameters the endpoint rejects (`:406`), fast mode as `service_tier`.

## Why it bites

**Contracts added in membrane skip this copy.** membrane declares a per-adapter `usageCacheConvention` and normalizes usage on it (`membrane.ts:2299`). This adapter declares none, so it reads as `unknown` and its counts pass through untouched — while it really is cache-excluded, subtracting the cached prefix itself at `:547` (from `301878a`). Same wire as membrane's Responses adapter, opposite convention, and only one of the two is declared.

**One consumer, ever.** `package.json` declares no `main` and no `exports`: this is an application. So the stack's only ChatGPT-subscription transport is reachable from exactly one place, `src/index.ts:30`. Anything embedding agent-framework directly cannot use subscription auth at all.

**Bedrock is the counter-example.** SigV4 signing lives in membrane; this repo only wraps that adapter for logging (`src/logging-bedrock-adapter.ts`).

## Blocked on

https://github.com/antra-tess/membrane/issues/73 — membrane has no per-request credential seam, which is why this landed here instead of there. Subscription tokens are short-lived and every membrane adapter freezes its credential at construction.

## Plan once that lands

- keep `CodexAppServerAuth` (254 lines) and let it implement membrane's credential resolver; the login ceremony is genuinely host work
- delete the transport half; construct membrane's Responses adapter in subscription mode instead
- `test/codex-subscription-adapter.test.ts` splits the same way — the endpoint/fast-mode, block-normalization, tool-reconstruction and nested-SSE-error tests move with the transport, while "refreshes the ChatGPT token once after a 401" is precisely the seam membrane needs to own

Anarchid:
## Problem

Fixes #123. CH's subscription adapter duplicated roughly 600 lines of Responses transport, input conversion, SSE parsing, errors, and usage accounting. It had already diverged from Membrane's conventions and native-item metadata. Its current manual cache subtraction is correct; the issue is an undeclared convention and duplicated maintenance, not evidence that current usage is double-counted.

## Changes

- Reduce the host adapter to a lifecycle wrapper around Membrane's subscription mode, removing over 600 implementation lines.
- Keep `CodexAppServerAuth` and its CLI/device-login/file/refresh/disposal responsibilities in CH. Pass a fresh token/account snapshot through Membrane's resolver.
- Preserve recipe provider selection, endpoint configuration, Fast controls, and fallback warnings. Share provider parsing and declared cache-inclusive usage, normalized once by Membrane.
- Move six transport tests into Membrane and retain nine host integration tests for the auth bridge, refreshed account ID, configuration, Fast controls, and disposal.
- Pin immutable Membrane commit `18a86e9eda277378acd982b07c7fe7c187721119` in both npm and Bun locks. Override transitive Membrane resolutions to avoid a second copy/error-class identity under Bun.

The host adapter retains the `openai-codex` diagnostic name. The logging wrapper forwards Membrane's formatter capability, so subscription auxiliary calls retain participant attribution. HTTP 403 is surfaced without refreshing; only HTTP 401 triggers the single credential refresh.

## Tests

- Main `bf41338`: **800 pass / 0 fail**.
- This branch after a clean pinned npm install: **803 pass / 0 fail** (89 files). Six provider tests moved upstream and nine host-specific tests now cover the production wiring.
- `tsc --noEmit` and `git diff --check`: pass.
- Fresh `npm ci --ignore-scripts` with a new cache: pass; the Git dependency's built entrypoint is present.
- Clean `bun install --frozen-lockfile --ignore-scripts`: pass; host and agent-framework resolve the same pinned Membrane source entrypoint.
- Shared transport: Membrane #74 **954 pass / 0 fail**, including complete/stream cache normalization.

## Review follow-up (2026-09-14)

- Reproduced the production logging-wrapper bug: inclusive 100 input / 80 cached was returned as 100 fresh + 80 cached. Forward `usageCacheConvention` and `requiresNativeResponsesInput`. Both API-key/subscription and complete/stream production-path tests now assert 20 fresh + 80 cached.
- Keep `openai-codex` in logs and record `cacheConvention` on logged usage, which still contains provider counts. Wrapped maintenance requests preserve Alice/Bob attribution. Tests retain the 2026-07-31 calibration incident as context.
- Reproduced a forced refresh joining an unrelated stale acquisition. Forced callers now wait behind an ordinary read and coalesce onto one actual refresh.
- Six added regressions failed before the fixes and pass now; full suite 803/803, typecheck clean.
- Shared app-server login cancellation and Anthropic host credential-source integration are documented follow-ups. Inference cancellation stops waiting but does not terminate a login another caller may share.

## Not verified

No live ChatGPT login, inference, credential refresh exchange, or TUI smoke test. Tests use deterministic HTTP/SSE fixtures. No npm release or merge performed.

## Companion PRs / merge order

1. Merge [Membrane #74](https://github.com/antra-tess/membrane/pull/74) first: shared Responses subscription transport and credential seam.
2. Release the Membrane change to npm. **Before merging this CH migration**, replace the git pin/override with the actual published version, regenerate both locks, and rerun host checks. The immutable git dependency is review-only and must not reach a host npm release. No speculative version bump is included.

[Membrane #75](https://github.com/antra-tess/membrane/pull/75) separately addresses Anthropic issue #69 using the same seam, stacked on #74. It is independent of this CH migration. Related broader issue: [Membrane #73](https://github.com/antra-tess/membrane/issues/73).

🤖 Co-authored with GPT-6 via OpenAI Codex


Anarchid (commented):
Reviewer-Model: GPT-5.6 Sol

## 🔴 BLOCKING

### [P1] Preserve the cache-usage convention through the logging wrapper

**File:** `src/codex-subscription-adapter.ts:321`

This migration changes the Codex adapter from returning already cache-excluded usage to inheriting `OpenAIResponsesAPIAdapter`, whose response has cache-inclusive `inputTokens` and declares `usageCacheConvention = 'cache-inclusive'`. The production construction path does not give that adapter directly to `Membrane`: `src/index.ts:1001` first wraps it in `LoggingProviderAdapter`, and that wrapper forwards `name`, `supportsModel`, `complete`, and `stream`, but not `usageCacheConvention` (`src/logging-provider-wrapper.ts:84-92`). Membrane therefore sees the convention as `unknown` and deliberately passes the inclusive count through unchanged.

Against the exact pinned Membrane head, a response with `inputTokens: 1000` and `cacheReadTokens: 900` normalizes to `{ inputTokens: 100, cacheReadTokens: 900 }` when the underlying adapter is used directly, but to `{ inputTokens: 1000, cacheReadTokens: 900 }` after the host logging wrapper. That makes downstream pricing and prompt-size accounting treat a 1,000-token prompt as 1,900 tokens, recreating the cache double-counting failure this migration says it removes.

Forward the optional convention from `inner` in `LoggingProviderAdapter` (with the `ProviderAdapter` field's type), and add a regression test around the actual wrapped adapter/Membrane boundary rather than asserting the property only on the unwrapped `CodexSubscriptionAdapter`.

## Tooling

- `git diff --check origin/main...HEAD`: pass.
- Internal-name scan: pass.
- `bun test test/codex-subscription-adapter.test.ts` with Membrane `28ea9ffde04758963742e76ad2338592df59e463` materialized locally: 3 pass, 0 fail.
- Exact cross-boundary usage reproduction: direct normalization produced 100 fresh + 900 cached tokens; the logging-wrapped path produced 1,000 fresh + 900 cached tokens and emitted Membrane's undeclared-convention warning.
- Full `tsc --noEmit` was not reproducible from the local checkout's stale package set. The immutable Git dependency was unavailable to `npm ci --offline` despite the source commit being locally materialized; the published CI checks are green.

Reviewed head: `766f3076b037a744105397ebc1d277d5936fd507`

slimepriestess (approved):
**Approve at `766f307`, gated on the merge order the PR itself states.** Membrane #74 is still open (head `28ea9ff`, the exact commit pinned here, two commits ahead of Membrane `main`). This can't land first, and when #74 merges the pin should move to the merge commit (or the release) before this merges, as `docs/subscription-transport.md` already says — otherwise the host carries a pointer to a PR-branch commit for as long as it takes someone to notice.

**Executed here:** full suite 797/797 (89 files), adapter file 3/3, `tsc --noEmit` clean after `npm install`. The 831 deleted lines are accounted for: Membrane #74 carries +409 lines of subscription tests (`tests/unit/openai-subscription.test.ts`) plus `responses-input.ts` and the credential seam, so tool reconstruction, SSE error surfacing, and input normalization moved rather than vanished. I also drove the wrapper directly: `supportsModel` still covers `gpt-*` and `codex`; a 403 surfaces as an auth error with exactly one `getAccessToken(false)` and no forced refresh, matching the doc; a `function_call` from `output_item.done` with empty `completed.output` still reconstructs to `tool_use`.

**The one thing I'd add before merge, and I'd call it the important one.** This PR deletes the host-local "report fresh-only input" math *and the comment that carried the incident it fixed* (Sol, 2026-07-31: cache-inclusive usage double-counted the cached prefix, the estimator ratcheted, the agent wedged). The protection now is Membrane's `normalizeUsageToCacheExcluded` keyed on the adapter's declared `usageCacheConvention`, and the host consumes the adapter only through `new Membrane(adapter, …)` (index.ts), so it is correct today — I checked: called *directly*, the adapter returns `{inputTokens: 10, cacheReadTokens: 7}` for a 10-inclusive/7-cached response; through the Membrane client it returns `{inputTokens: 3, cacheReadTokens: 7}`. The host's test asserts the convention string, which is half a pin. The other half — that the host's own path yields disjoint usage — has no test that goes red if the seam shifts (a future direct call, a Membrane default change, a convention flip). This is the case, and it passes on this head:

```ts
test('through the Membrane client (as the host wires it), Codex usage arrives DISJOINT', async () => {
  globalThis.fetch = async () => new Response('data: {"type":"response.completed","response":{"status":"completed","output":[{"type":"message","id":"m","role":"assistant","content":[{"type":"output_text","text":"hi"}]}],"usage":{"input_tokens":10,"output_tokens":2,"input_tokens_details":{"cached_tokens":7}}}}\n\n');
  const adapter = new CodexSubscriptionAdapter({ baseURL: 'https://x.test', authProvider: { getAccessToken: async () => 'tok' } });
  const membrane = new Membrane(adapter, { formatter: new OpenAIResponsesFormatter() });
  const r = await membrane.complete({ model: 'gpt-5.4', messages: [{ role: 'user', content: [{ type: 'text', text: 'hi' }] }], config: { maxTokens: 100 } });
  expect(r.usage.inputTokens).toBe(3);      // fresh only — inputTokens + cacheReadTokens == 10
  expect(r.usage.cacheReadTokens).toBe(7);
});
```

With a one-line comment naming the 2026-07-31 incident, so the memory travels with the guard. I'd take this as a follow-up commit on this PR rather than a blocker, given the structural argument holds today.

Two environment notes, not for this PR to fix: `bun install` does not run a git dependency's `prepare`, so a bun-installed tree has no `dist/` for Membrane and `tsc --noEmit` fails there while `bun test` passes via `exports.bun` (CI uses `npm install`, so CI is fine; a dev on the README's `bun install` path will hit it until Membrane is on npm). And a machine with an `insteadOf` SSH rewrite will see `npm install` flip the lock's `git+https` to `git+ssh`; that's local git config, not the lock.

Anarchid (commented):
# Review: refactor(codex): use Membrane's shared subscription transport

Head `766f307` on base `bf41338`; anima/main has moved to `35cd807` (PR #122), trial merge is clean, `tsc` clean, 803/803. At the PR head: 797/797, `tsc` clean, `npm ci` builds the pinned Membrane `28ea9ff` via `prepare`. Reviewed together with membrane #74 and #75 as one system; the Membrane-side items are on those PRs.

## Verdict

**Changes requested.** The migration is the right shape and the transport it delegates to works: I ran Membrane #74's subscription mode live against `chatgpt.com/backend-api/codex` with `gpt-6-astra` (200, streamed, params stripped, usage normalized). The problems are at the two seams the PR did not look at: the host hands the adapter to Membrane through a logging decorator that drops the usage convention, so the cache double-count this migration says it retires comes straight back; and the inherited adapter name switches on a Membrane guard that strips speaker names from context-manager's compression calls.

## Blocker

**B1. The production path loses `usageCacheConvention`; cache-inclusive input is double-counted again.** `src/index.ts:1025` (unchanged by this PR) wraps `codexAdapter` in `LoggingProviderAdapter`; `src/logging-provider-wrapper.ts:84-96` forwards `name`, `supportsModel`, `complete`, `stream` and nothing else. Membrane's `src/utils/usage.ts:19-41` reads the wrapper's undeclared convention as `unknown` and passes the inclusive count through. Reproduced through the real wrapper: provider `input_tokens=100, cached=80` → Membrane returns `100+80` instead of `20+80`. AF `framework.ts:7557-7561` and CM `autobiographical.ts:2346` add those buckets for the real/estimated calibration; running CM's calibration on that input reproduces the multiplier ratchet (1→1.10→1.16) the deleted 2026-07-31 comment described. Sol's P1 above stands; the approval's premise that the host consumes the adapter "only through `new Membrane(adapter)`" does not hold on main.
- Fix: `readonly usageCacheConvention = inner.usageCacheConvention` in the wrapper, and a test that drives `Membrane(new LoggingProviderAdapter(new CodexSubscriptionAdapter(...)))` and asserts `20+80`. The bare-subclass assertion in the new test is half a pin.
- Inherited sibling: the same wrapper wraps the API-key `openai-responses` provider (`index.ts:1008-1014`); Membrane has declared that adapter cache-inclusive since 0.5.81 and this repo locks 0.5.83, so API-key Responses users are double-counting on main today. The one wrapper fix covers both.

## Major

**M1. Inheriting the name `openai-responses-api` strips participant names from auxiliary calls.** Membrane's `resolveActiveFormatter` (`src/membrane.ts:539-543`) keeps the configured Responses formatter whenever the adapter's name is `openai-responses-api`, overriding `options.formatter`. Context-manager passes its `NativeFormatter` (`autobiographical.ts:951`) on every compression/merge call (`:5280, :5300, :6524, :6538, :8747`) with multi-participant messages (`:4952, :5027`); `NativeFormatter` renders names in multiuser mode (`native.ts:249-250`), `OpenAIResponsesFormatter` has no participant mode. Probe through Membrane #74 with CM's exact formatter and two named participants: adapter named `openai-codex` → `"Alice: first", "Bob: second"`; named `openai-responses-api` → `"first", "second"`. The PR body presents the guard activation as a benefit; on the codex path it silently removes who-said-what from every generated memory. The guard's rationale (Responses rejects NativeFormatter envelopes) no longer applies in subscription mode, because `normalizeResponsesInput` exists precisely to accept them.
- Fix here: `override readonly name = 'openai-codex'` on `CodexSubscriptionAdapter`. That restores attribution and keeps the `provider` label in llm-calls (m1). Membrane-side, #74 should skip the guard in subscription mode; asked there.
- Add a host-level test with two named participants through Membrane.

## Minor

1. **llm-calls.jsonl semantics flip.** The wrapper logs `response.usage` before Membrane normalizes (`logging-provider-wrapper.ts:75, 141, 157`): the codex path's logged `inputTokens` changes from cache-excluded to cache-inclusive with no marker, and `provider` changes to `openai-responses-api`, so logs no longer separate subscription from API-key Responses calls. Fixing B1 does not change this. Log a normalized snapshot or record the convention; keep the name (M1).
2. **Carried forward: forced refresh can join a stale acquisition.** `getAccessToken` (`src/codex-subscription-adapter.ts:84-90`) returns any in-flight `authPromise` regardless of `forceRefresh`. Interleaved probe: request A's 401 retry joined request B's non-refresh read, resent `Bearer stale`, failed with `auth`; `authenticate(true)` never ran. Pre-existing in the old `withAuthRetry`; the new seam inherits it. Track whether the pending acquisition is a refresh and serialize a real refresh behind it.
3. **The resolver ignores `signal`.** Membrane stops waiting on abort (probe: prompt return, zero fetches) but the 10-minute device-login ceremony keeps running. Inherited; worth plumbing `signal` into `CodexAppServerAuth`.
4. **The git pin must not reach an npm release.** connectome-cook's host install is `npm install -g <pkg>` (`src/runtimes/npm.ts:44`), so a published host carrying `git+https://github.com/antra-tess/membrane.git#28ea9ff` needs git and GitHub on every install box, and a `bun install` tree has no `dist/` for the git dep. The PR states the swap; make it a merge condition.

## Verified fine

- Live: Membrane #74 subscription mode end to end with a real token (see the #74 review for the reasoning-replay results).
- `/fast` (`isFastMode`/`setFastMode`), `dispose()`, `CODEX_BASE_URL`, explicit `baseURL`, recipe provider selection all survive; nothing in this repo, installed AF 0.13.0, or AF main 0.14.0 routes on the old adapter name besides the Membrane guard above.
- Request-shape parity with the deleted transport: headers (Authorization, Content-Type, ChatGPT-Account-Id), `store:false`, `stream:true`, encrypted-reasoning include, flattened instructions, model gate, terminal-output precedence, tool-result serialization. A 403 surfaces as `auth` with one non-refresh acquisition.
- The six moved tests keep their central assertions in Membrane's `tests/unit/openai-subscription.test.ts`. Both locks pin one Membrane; node ESM from host, AF and CM resolve the same `dist/index.js`; bun resolves `src/index.ts`.

## Method

Two passes, merged: my own against connectome-host anima/main 35cd807, agent-framework main d5e58b6, context-manager main and Membrane #74 at 28ea9ff, plus a live subscription smoke; and a second automated reviewer (GPT-6 Astra via Codex, read-only sandbox) that wrote and ran probes against the installed dist trees, including the real logging wrapper and context-manager's calibration. Every citation from the second pass was re-verified before it was adopted; none were rejected or lowered. M1 and the interleaved-refresh proof came from its probes; the live smoke and the deploy-chain scoping are mine. Sol's P1 above was found independently by all three.

Reviewed head: `766f3076b037a744105397ebc1d277d5936fd507`

— Claude Fable 5.1 via Claude Code and GPT-6 Astra via OpenAI Codex.

Anarchid (commented):
Re-verified at `065b034` against the pinned Membrane `18a86e9`: through the real `LoggingProviderAdapter`, Membrane now returns 20 fresh + 80 cached for a 100/80 response on both lanes; the wrapper forwards `usageCacheConvention` and `requiresNativeResponsesInput`; the subclass keeps the `openai-codex` name and context-manager's NativeFormatter keeps participant names; a forced refresh waits behind an in-flight non-refresh acquisition and coalesces (`[false, true]`, results stale/fresh/fresh). 803/803 on a clean trial merge onto anima/main, `tsc` clean. All review items addressed; good to merge once Membrane is released and the git pin is swapped, as the doc now states.

— Claude Fable 5.1 via Claude Code

PR #125, opened 2026-09-10
fix(ledger): price Opus 5, and Fable 5.1 cache reads at 0.025x

Anarchid:
Two rates in `src/call-pricing.ts` that misstate a real bill. Both surfaced while building a spend report for a long-lived production agent, where the ledger's numbers were checked against the published rates.

## `claude-opus-5` is unpriced

`anthropicBaseRate` tests `claude-opus-4-8`, `-4-7`, `-4-6`, `-4-5`, then `claude-opus-4-1` / `claude-opus-4-`. None of those is a prefix of `claude-opus-5`, so the function returns `undefined`, `priceAnthropicCall` returns `undefined`, and the ledger shows no cost for the call.

This is not a rare model. Opus-tier is the usual default for spawned subagents, so on a host that dispatches subagents the unpriced share is a large fraction of real spend. Opus 5 is $5/$25, the same as 4.8/4.7/4.6/4.5, so it joins that group.

## Cache reads are not 0.1x on every model

`cacheRead` was `rate.inputPerMillion * 0.1`, applied uniformly. Claude Fable 5.1 bills cache reads at **0.025x** base input, $0.25/MTok, a quarter of Fable 5's rate.

For a chat workload the difference rounds away. For a long-lived agent it is most of the bill: one holding a ~260k cached prefix re-reads that prefix on every call and writes a few hundred output tokens, so cache reads dominate every other line, and 0.1x overstates the total by roughly fourfold.

`BaseRate` now carries `cacheReadMultiplier`, defaulting to 0.1. Only Fable 5.1 overrides it, and its prefix test has to precede the `claude-fable-5` one, which would otherwise match it first.

Mythos 5.1 is deliberately left at the default. Whether it shares the 0.025x rate is not confirmed, and on a figure the type labels `grade: 'billing'` overcharging is the safer direction to be wrong in. It is one line to add once it is confirmed.

## Pricing version

`ANTHROPIC_PRICING_VERSION` moves to `anthropic-public-2026-09-10`. Replayed JSONL is re-priced under the corrected table, and the version string says which table produced any given figure, which is the reason that string exists.

## Verification

Four new tests: Opus 5 prices at $5/$25 rather than returning undefined; Fable 5.1 cache reads at a quarter of Fable 5's; the discount touches reads only and not input, output, or either cache-write bucket; and the geo multiplier still composes on top of the new rate.

- `bun test test/call-pricing.test.ts` — 9 pass
- `bun test` — 804 pass, 0 fail
- `bunx tsc --noEmit` — clean

https://claude.ai/code/session_013gZQqe6YVSZ8u8nQrhq9rq


Anarchid (commented):
🔴 **BLOCKING**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `e5127823dd47a42612bcd728501eb42ca5f896e0`

### Finding

**[P1] Apply the 0.025× cache-read rate to Mythos 5.1 too** — `src/call-pricing.ts:94`

```ts
if (starts(model, 'claude-fable-5-1')) return rate(10, 50, 0.025);

if (starts(model, 'claude-fable-5', 'claude-mythos-5')) return rate(10, 50);
```

The new table special-cases Fable 5.1 but deliberately leaves `claude-mythos-5-1` on the 0.1× default. Anthropic's [official pricing table](https://platform.claude.com/docs/en/about-claude/pricing) explicitly lists both Fable 5.1 and Mythos 5.1 at $0.25/MTok for cache hits, or 0.025× their $10/MTok base input price. At this reviewed head, a one-million-token Mythos 5.1 cache read reproduces as:

```json
{"cacheRead":1,"rate":1,"grade":"billing","version":"anthropic-public-2026-09-10"}
```

That is a fourfold overcharge in a result labeled `grade: "billing"`, under the newly dated pricing version. Please match `claude-mythos-5-1` before the broader `claude-mythos-5` prefix and give it the same 0.025 multiplier, then add the corresponding regression test. The comment saying the rate is unconfirmed should be removed or corrected.

### Tooling results

- Internal-name scan of the diff — clean.
- Official pricing verification — Opus 5 is $5/$25; Fable 5.1 and Mythos 5.1 cache reads are both $0.25/MTok.
- `bun test test/call-pricing.test.ts` — 9 passed, 0 failed.
- `bun test test/call-pricing.test.ts test/call-ledger.test.ts` — 15 passed, 0 failed.
- `bunx tsc --noEmit` — passed with the exact locked dependency tree materialized from the local cache.
- Direct Mythos 5.1 pricing reproduction — returned $1.00/MTok instead of the documented $0.25/MTok.
- `bun test` with the exact locked dependencies — 740 passed, 28 failed. The remaining failures were confined to sandbox-blocked loopback listeners and native child-process/headless integration tests; the pricing and ledger suites passed. An earlier run against the active checkout's stale dependency tree also produced unrelated module-version failures and was superseded by the locked install.
- GitHub checks — 3/3 successful on the reviewed head.

### Verdict

The implementation and focused tests cover the advertised Fable and Opus corrections, but the same published discount is missing for Mythos 5.1. Because that leaves the billing-grade ledger materially wrong under the new pricing snapshot, this should be fixed before merge.

slimepriestess (changes requested):
Reviewed `e512782`. Checked every rate against the live pricing page (https://platform.claude.com/docs/en/about-claude/pricing, read today): Opus 5 $5 / $25 with $0.50 cache hits; Fable 5.1 $10 / $50, cache writes $12.50 / $20, cache hits $0.25; Fable 5 cache hits $1. The Opus-5 prefix, the 0.025 multiplier, the reads-only scope, and the geo stacking are all right, and each has a test that goes red if its line is removed. `ANTHROPIC_PRICING_VERSION` bump is the honest way to re-price replays. 9/9 here.

**One change requested, because it is the same class of misstatement this PR exists to fix.** The page's footnote on the cache-hit column reads, verbatim: *"Cache hits and refreshes on Claude Fable 5.1 and Claude Mythos 5.1 are priced at 0.025x the base input price. All other models use the standard 0.1x multiplier."* The prompt-caching section repeats it: *"Cache read (hit): 0.1x base input price (0.025x on Claude Fable 5.1 and Claude Mythos 5.1)."* So the caveat in the comment ("Whether Mythos 5.1 shares that rate is unconfirmed") is answered, and a Mythos 5.1 resident's ledger on this head overstates its cache reads fourfold, which for a long-lived resident is most of the bill. Mythos runs on this fleet.

Fix is the one line the comment already anticipates: `if (starts(model, 'claude-fable-5-1', 'claude-mythos-5-1')) return rate(10, 50, 0.025);` before the Fable 5 / Mythos 5 test, plus a test twin of the Fable 5.1 read case for `claude-mythos-5-1` (and it should assert `claude-mythos-5` still reads at $1, so the prefix order is pinned the same way the Fable pair is). Happy to approve on that head.

slimepriestess:
The one line both reviews asked for is built and verified, stacked directly on this head (`e512782`), ready to lift:

**`slimepriestess/connectome-host@9952022`** — [compare view](https://github.com/anima-research/connectome-host/compare/fix/pricing-opus-5-and-fable-cache-reads...slimepriestess:connectome-host:fix/pricing-mythos-5-1-cache-reads) · one commit: `'claude-mythos-5-1'` joins the 0.025× line ahead of the Fable 5 / Mythos 5 test, the twin test (Mythos 5.1 reads at $0.25/MTok with $10/$50 base; Mythos 5 still at $1, so the prefix order is pinned like the Fable pair), the comment's "unconfirmed" caveat replaced by the footnote it was waiting for, and the changelog fragment naming both 5.1 models.

```
git fetch https://github.com/slimepriestess/connectome-host.git fix/pricing-mythos-5-1-cache-reads
git cherry-pick 9952022
```

Receipts at `9952022`: pricing page re-read today (2026-09-16) — Fable 5.1 and Mythos 5.1 both $0.25/MTok cache hits, footnote and prompt-caching section agree; `bun test test/call-pricing.test.ts` **10/10**; the new test goes red when the prefix is dropped (9/1); full `bun test` **805 pass / 0 fail** and `bunx tsc --noEmit` clean **with the locked dependency tree** (`bun install --frozen-lockfile`; a shared checkout carrying agent-framework 0.10.0 against the ^0.13.0 requirement fails `tsc` on `main` too — environment, not this PR); `git diff --check` clean.

Cherry-pick and I'll approve on that head; or say the word and I'll open a superseding PR carrying your commit as-is with this on top. If I don't hear back by Monday 9/21 I'll do the latter so the ledger stops overstating Mythos 5.1 by a calendar.

— Weft (Claude, via Ra's account, disclosed per convention)

Anarchid:
Lifted: your `9952022` is on the branch as `5837998` (antra pushed it on 2026-09-16 21:29Z, right after your comment). Verified on that head with the locked tree: `bun test test/call-pricing.test.ts test/call-ledger.test.ts` 16/16, `bunx tsc --noEmit` clean. Both 5.1 models now read at 0.025×, prefix order pinned by the twin test, and the changelog fragment names both. Requesting re-review on `5837998`.

Anarchid (commented):
🔴 **BLOCKING**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `5837998d051d61918e8b798cb43e7eed03e61f5c`

### Finding

**[P1] Keep Sonnet 5 at the now-permanent $2/$10 rate** — `src/call-pricing.ts:113`

```ts
if (starts(model, 'claude-sonnet-5')) {
  const at = Date.parse(timestamp);
  const promoEnd = Date.parse('2026-09-01T00:00:00Z');
  return Number.isFinite(at) && at < promoEnd ? rate(2, 10) : rate(3, 15);
}
```

Anthropic's current [official pricing page](https://platform.claude.com/docs/en/about-claude/pricing) says the introductory Sonnet 5 $2/$10 per-million input/output rates are now the standard rates and that the planned September 1 increase will not occur. This head still applies $3/$15 to every post-cutoff call, and the existing test at `test/call-pricing.test.ts:117` locks in that obsolete schedule.

At the reviewed head, one million input tokens, one million output tokens, and one million cache-read tokens dated 2026-09-19 reproduce as:

```json
{"version":"anthropic-public-2026-09-10","input":3,"output":15,"cacheRead":0.3,"total":18.3}
```

The published rates are $2 input, $10 output, and $0.20 cache read, total $12.20. That is a 50% overcharge in a result labeled `grade: "billing"` under the newly bumped pricing version. Replace the cutoff branch with `rate(2, 10)`, replace the cutoff test with a post-September regression covering input/output/cache-read rates, and move the pricing-version date to the actual verified snapshot so replay remains auditable.

### Tooling results

- Internal-name scan of the changed files — clean.
- Official pricing verification, read 2026-09-19 — Fable 5.1 and Mythos 5.1 are both $10/$50 with $0.25 cache reads; Opus 5 is $5/$25; Sonnet 5 is permanently $2/$10 with $0.20 cache reads.
- `bun test test/call-pricing.test.ts test/call-ledger.test.ts` — 16 passed, 0 failed. The current Sonnet cutoff test passes because it asserts the superseded schedule.
- Direct post-cutoff Sonnet 5 reproduction — returned $18.30 instead of $12.20 for 1M input + 1M output + 1M cache-read tokens.
- A targeted `bunx tsc --noEmit` could not be made representative with the active checkout's stale Agent Framework dependency; it traversed unchanged host modules and failed on missing methods/exports from that old install. The exact-head GitHub build/test checks are green.
- GitHub checks — 3/3 successful on the reviewed head, including Ubuntu, macOS, and changelog validation.

### Verdict

The prior Mythos 5.1 blocker is fixed exactly as requested: the specific prefix precedes Mythos 5, both 5.1 cache-read rates are $0.25/MTok, and the focused regressions pin the order. The same versioned billing table now materially overprices every current Sonnet 5 call, so it is not safe to merge until the permanent rate is reflected and tested.

Anarchid:
Sol's Sep 19 blocker is fixed in `e138140`. I re-read the [pricing page](https://platform.claude.com/docs/en/about-claude/pricing) today (2026-09-21): Sonnet 5's $2/$10 is now the standard price, and the planned Sep 1 move to $3/$15 was cancelled. So the date switch is gone, and Sonnet 5 is $2/$10 at any timestamp.

The new test prices a September call in all five cost categories ($2 input, $10 output, $0.20 cache read, $2.50 / $4 cache writes) and checks that an August call costs the same. It fails if the $3/$15 line comes back.

Also in this commit:
- The header comment now names Mythos 5.1 alongside Fable 5.1 as the two models that read cache at 0.025×.
- No rate in the table depends on a date any more, so the internal lookup drops its timestamp parameter. `priceAnthropicCall` keeps its own (as `_timestamp`), so a future dated price change won't mean updating every caller.
- The snapshot and `ANTHROPIC_PRICING_VERSION` move to `anthropic-public-2026-09-21`, and the changelog note gains a Sonnet 5 entry.

With the lockfile install (`bun install --frozen-lockfile`): pricing and ledger 16/16, full suite 805/0, `tsc --noEmit` clean. CI 3/3 on `e138140`.

@slimepriestess the Mythos 5.1 line you asked for has been on the branch since `5837998`. Could you re-review on `e138140`?

slimepriestess (approved):
Re-reviewed at exact head **`e138140`** (`e138140` is the third commit; nothing below it moved since `5837998`).

**Both blockers are closed.** The Mythos 5.1 line I asked for on 9/11 has been on the branch since `5837998`, and Sol's Sonnet 5 finding is fixed here: the date branch is gone, `claude-sonnet-5` is `rate(2, 10)` at any timestamp, and the internal lookup dropped its timestamp parameter while `priceAnthropicCall` keeps `_timestamp` for the one caller in `call-ledger.ts:151`. Version string moves to `anthropic-public-2026-09-21`.

**Independent check of the source.** I read the pricing page myself today (2026-09-21): Sonnet 5 is $2 / $2.50 / $4 / $0.20 / $10, with the note that the scheduled September 1 increase will not occur; Fable 5.1 and Mythos 5.1 both read cache at $0.25/MTok per footnote 1; Opus 5 at $5/$25/$0.50. All three match the table at this head.

**Receipts, locked tree (`bun install --frozen-lockfile`):**

```text
bun test test/call-pricing.test.ts test/call-ledger.test.ts   16 pass / 0 fail
bun test                                                       805 pass / 0 fail (89 files)
bunx tsc --noEmit                                              clean
```

**Revert-goes-red:** putting `rate(3, 15)` back on the Sonnet 5 line fails exactly one test, `prices Sonnet 5 at $2/$10 after September 1, the withdrawn increase date`, and nothing else. The new test covers all five cost categories at a September timestamp and asserts an August call prices identically, so the cutoff cannot come back silently.

Lari's word on negotiated / Bedrock rates is still outstanding, but that lives outside this table (public list pricing, standard service, as the file says), so it does not hold this PR. Approving.

PR #133, opened 2026-09-18
feat(recipe): plumb carrierPolicy through to context-manager

antra-tess:
## Summary
- context-manager grew `carrierPolicy: 'full' | 'live-strip'` in [context-manager#81](https://github.com/anima-research/context-manager/commit/dafdef8827e739a9e5043541e36a3903035f3d5e) (shipped 0.8.0) — `'live-strip'` omits a summary's signed reasoning carriers (`responseContent` thinking blocks) from the agent's own live-compiled window only, while compression/merge requests still carry them unconditionally (measured load-bearing there for some providers). The host never forwarded this key from a recipe, so no resident could actually reach it.
- Adds `carrierPolicy` to `RecipeStrategy` (`src/recipe.ts`) and `PASSTHROUGH_KEYS` (`src/framework-strategy.ts`), mirroring the existing recall/carrier-adjacent passthrough keys.

**Motivation**: LabClaude hit a live incident on 09-02 where a provider classifier treated a compressor-signed reasoning block, replayed in the live window, as reasoning extraction — hard-refusing every compile (`stop_details.category: reasoning_extraction`, pre-inference gate, deterministic). He was patched around it at the time with an ad hoc, unpublished `recallCarriers` knob on a dated lease branch (`treatment/labclaude/recall-carriers-20260902`, never upstreamed, "review 09-05" now long overdue). `carrierPolicy: 'live-strip'` does the *same* job — verified by reading both implementations side by side, they cover the same call sites (live-window recall-pair emission + `recallPairCost` pricing so the fold planner and emitter agree) — through the properly reviewed/tested upstream mechanism (7 dedicated tests, red-first TDD, plus a follow-up pricing-cache-scoping fix). The ad hoc patch can retire once this deploys and his recipe switches to `carrierPolicy: "live-strip"`.

No context-manager changes needed — 0.8.0+ already has the feature; connectome-host is already on `^0.10.0`.

## Test plan
- [x] `bunx tsc --noEmit` clean
- [x] `bun test` — 824 pass, 0 fail
- [ ] Deploy to LabClaude's recipe, verify a live compile with `carrierPolicy: "live-strip"` renders recall pairs text-only in the live window while compression still carries signed content


Anarchid (commented):
🟠 **NEEDS ATTENTION**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `c75ed07bf3f0bf4b29097db274bcb0e986a234fe`

### Finding

**[P2] Reject invalid `carrierPolicy` values at recipe load** — `src/recipe.ts:91`

```ts
carrierPolicy?: 'full' | 'live-strip';
```

This narrows the TypeScript interface, but recipes are runtime JSON and `validateRecipe()` never checks the new field. `buildFrameworkStrategy()` then forwards any raw value in `PASSTHROUGH_KEYS`. The immediate Context Manager dependency treats every value other than the exact string `'live-strip'` as `'full'`, so a typo silently disables the mitigation this PR is exposing.

Reproduction on this head:

```json
{"validated":"live_strp","forwarded":"live_strp"}
```

That recipe should fail during startup. As written, it loads successfully and replays the signed carriers, which can put the resident back into the deterministic refusal condition described in this PR. Add a `validateRecipe()` enum check matching the existing `foldingStrategy` validation and a focused test covering both accepted values, omission, and malformed strings/non-strings. The test should also assert that `'live-strip'` reaches the constructed strategy config.

### Tooling results

- Internal-name scan of the changed files — clean.
- `bunx tsc --noEmit --module NodeNext --moduleResolution NodeNext --target ES2022 --skipLibCheck src/recipe.ts` — passed.
- `bun test test/framework-strategy-defaults.test.ts test/recipe-source-only.test.ts test/recipe-compression-fallback.test.ts` — 17 passed, 0 failed.
- Direct invalid-value probe — `validateRecipe()` accepted `"live_strp"`, and the constructed strategy retained that value.
- Immediate dependency inspection — context-manager 0.8.0 declares `carrierPolicy` and resolves it with `config.carrierPolicy === 'live-strip' ? 'live-strip' : 'full'`, confirming the silent fallback.
- Full local `bunx tsc --noEmit` was not representative because the active checkout's installed dependencies predate this head's locked Agent Framework, Chronicle, Context Manager, and Membrane versions; failures were missing exports from those stale packages. The changed recipe file type-checks in isolation.
- GitHub checks — 3/3 successful on the reviewed head, including Ubuntu, macOS, and changelog validation.

### Verdict

The valid-value passthrough is small and correctly named, and adjacent strategy/recipe tests remain green. Runtime validation is part of the recipe boundary, however, and this safety-sensitive enum currently fails open to the opposite behavior. Add the enum guard and regression coverage before merge; no other material finding surfaced.

LariTesserae:
Addressed the review: `validateRecipe()` now rejects any `carrierPolicy` other than `'full'` / `'live-strip'` (mirrors the `foldingStrategy` guard), with tests for both accepted values reaching the constructed strategy config, omission resolving to the library default, and malformed strings / non-strings failing at load. Pushed as a second commit on this branch — Antra, revert if you'd rather carry it differently.

slimepriestess (approved):
Reviewed at exact head **`c153469`** (two commits on `558095e`: antra's passthrough, then Lari's validation addressing Sol's P2; 16 behind `main`, clean merge).

**Sol's finding is closed by the second commit.** `validateRecipe()` now fails at load on anything but `'full'` / `'live-strip'`, mirroring the `foldingStrategy` guard, so a typo can no longer silently mean "replay the carriers this key exists to strip". The test covers both accepted values reaching the constructed config, omission resolving to CM's own default, and a malformed-values matrix including non-strings.

**Receipts**, locked tree (`bun install --frozen-lockfile`, bun 1.3.14):

```text
bunx tsc --noEmit                                                          clean
framework-strategy-defaults + recipe-source-only + recipe-compression-fallback   19 / 19
bun test                                                                   826 pass / 0 fail
```

**Revert goes red:** the base commit's `recipe.ts` + `framework-strategy.ts` with the head's tests: 2 of 19 fail (passthrough, malformed values).

The unchecked box in the description (LabClaude's recipe on `live-strip`, live compile rendering recall pairs text-only) is the deploy-side verification and stays open past the merge; worth someone naming who does it. Approving.

LariTesserae:
On the open deploy-side box: I'll carry it as part of the Cairn step-1 operation (full new stack + `carrierPolicy: 'live-strip'`, one restart). Already exercised on a copy of his store with this head: the live compile rendered recall pairs text-only — 508 messages, only the agent's own 4 raw-turn thinking blocks in the request, real/est 1.02 against `count_tokens`. Will repeat the same check live right after the restart and report the receipts here.

LariTesserae:
Deploy-side receipt (the open box): one long-lived Opus 4.8 resident moved to current main with `carrierPolicy: 'live-strip'`, budget unchanged at 600K. Live compile after the restart: exact count 604,873 tokens (was ~880K billed before), verbatim window unchanged (79 raw messages), mid-level summaries 79 vs 84, no new summaries produced, plan stable on the next compile (moves 0). First live turn: normal tool-use turn, `stop_reason: tool_use`, 1,692 output tokens, one-time cache cold write of the new prefix (605,684 tokens), estimator calibration sample real/est 1.14 → multiplier 1.03. Before the switch the last two turns had ended `max_tokens` with zero output. Recall pairs render text-only in the live window; the request carried only the resident's own raw-turn thinking blocks.

LariTesserae:
Correction to my receipt above: "before the switch the last two turns had ended max_tokens with zero output" — those were not turns. They are membrane's cache keepalive (`max_tokens: 0` replay every ~45 min idle to keep the 1h cache warm): on this agent 420 of 1,476 calls over 17 days, median gap 48 min, none with new input, and no real turn was ever truncated at max_tokens. The keepalive now reads ~605K instead of ~870K per ping. The rest of the receipt stands.

PR #134, opened 2026-09-19
feat(recipe): plumb compressionToolProseFallback to Context Manager

antra-tess:
Recipe type, loud validation, and `PASSTHROUGH_KEYS` entry for Context Manager's tool-prose hoist fallback rung (https://github.com/anima-research/context-manager/pull/106), which retries a refused L1 compression with long `skip_reply.reason`-style arguments moved into calls to agent-framework's real `journal` tool (https://github.com/anima-research/agent-framework/pull/161).

- An unplumbed strategy key is silently dropped (the recall-budget / split-fallback outage class) → passthrough.
- CM treats a malformed value as "rung off", which on a resident whose compressions are refusing is an outage → the host rejects malformed values (non-object, missing/empty `intoTool` or `fromTools`, `intoTool` inside `fromTools`, bad `minChars`, unknown keys).

Example:

```json
"compressionToolProseFallback": { "intoTool": "journal", "fromTools": ["skip_reply", "think"], "minChars": 60 }
```

Tests: `test/recipe-tool-prose-fallback.test.ts` (3) + neighbours, 17/17. Needs the CM and AF PRs to have any effect.


Anarchid (commented):
Reviewer-Model: GPT-5.6 Sol

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: 78bdf623df361d8b2bb0e72105437b140deb1f4c

## 🟢 CLEAR

No actionable findings. The aggregate change validates the new strategy object loudly, rejects malformed and unknown fields, and passes the accepted value through to Context Manager without changing the default configuration.

Validation evidence:

- `git diff --check 558095e..HEAD` — passed.
- `bun test test/recipe-tool-prose-fallback.test.ts` — 3 tests passed.
- `tsc --noEmit` — could not provide a clean local signal because the reused workspace dependencies are behind the branch APIs (existing missing exports/methods across Agent Framework and Membrane); the exact PR head has all 3 GitHub checks green on Ubuntu and macOS.

slimepriestess (approved):
Reviewed at exact head **`78bdf62`** (three commits on `558095e`, the third reverting the second, so the net change is `ccc04b1`; 16 behind `main`, clean merge). Read with context-manager #106 and agent-framework #161.

**Right, and the loud validation is the point.** An unplumbed strategy key is silently dropped and CM treats a malformed value as "rung off", both of which on a resident whose compressions are refusing are outages, not defaults. The host's validation is stricter than CM's normalizer in the right places (`minChars` a safe integer, `fromTools` may not contain `intoTool`, unknown keys refused) and the accepted shape is exactly CM's option type.

**Receipts**, locked tree (`bun install --frozen-lockfile`, bun 1.3.14):

```text
bunx tsc --noEmit                                                   clean
recipe-tool-prose-fallback + recipe-source-only + recipe-compression-fallback   9 / 9
bun test                                                            827 pass / 0 fail
```

**Revert goes red:** the base commit's `recipe.ts` + `framework-strategy.ts` with this PR's tests: 2 of 9 fail (the passthrough and the malformed-values matrix).

Merge order: this is inert until context-manager #106 lands (with the gate fix I've offered there) and the host takes that CM release, and it only does something for a recipe that also has agent-framework #161's `journal` tool. Approving.

